PR #4, opened 2017-08-24
Fix server-side eslint

alex-whitney:
These changes might be a bit aggressive, as it looks like a lot of this code is still early in development. However, there are a few typos that eslint flagged, which I've addressed.


gryffon:
I've looked it over, and I'm not seeing anything that strikes me as bad. A fair amount is still in an active state, but it's easy enough to bring back.

PR #5, opened 2017-08-25
Mulligan

SnarkAttack:
Added mulligan for both conflict and dynasty decks. Mulligans alternate between players, as it's not possible to have them be simultaneous with the current select cards functionality. Added chat logging for mulligan/keep choices and number of cards mulliganed if applicable. Deleted all old mulligan files.

Made a few small changes elsewhere, including:
- Added holdings to default selectable cards in selectcardprompt.js
- Set basecard.js to default isDynasty (and other similar values) to false
- In fillProvinces() (player.js), I had to pull out the actual array of the getSourceList() result by using the ._wrapped attribute. fillProvinces was failing every time because it was passing in an array or cards as the parameter for the nested function instead of individual cards. I didn't poke around to see if this was an issue elsewhere, but it was a pain in the ass to deal with so I figured I'd document it now.


elstob:
I imagine having simultaneous mulligans is going to be pretty important in the future as having the information of if your opponent was digging for a card hard or not may effect your choice to do the same. Is it at least possible to hide details of how many cards were changed until both players have changed?

gryffon:
I don't think a lot of people have been doing it, but per the RRG mulligans are actually in player order:

◊ Players mulligan (or pass the opportunity to do so) in player
order. If the first player passes an opportunity to mulligan,
that player cannot change his or her mind and then decide
to mulligan during that step after seeing the opponent’s
decision.

Provinces first, then hand.

elstob:
Ah cool, that makes things a whole lot easier then 😄

SnarkAttack:
I looked through the learn to play, but not the rules reference, so lucky more than anything here. I did think it makes sense for tourney play that there would be a defined order though, otherwise people would sit there waiting for their opponent to mulligan for this info.

PR #36, opened 2017-09-09
Fix for reappearing attachments

jeremylarner:
I'm afraid I'm not familiar with _.js so this may be a horrible fix - please fix it if there's a more elegant way of doing it.  Seems to fix all the attachment issues though.


alex-whitney:
There do not appear to be any changes in this PR.

jeremylarner:
Sorry, I'm still learning my way around github.  The change that was supposed to be here is the one your commented on regarding this.removeAttachments in the other PR.

PR #38, opened 2017-09-09
Deck popup fix

jeremylarner:
Dynasty and Conflict Decks use legacy CSS
Fix for top card showing as facedown in Popups

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -960,7 +960,7 @@ class Player extends Spectator {
             card.attachments.each(attachment => {
                 this.removeAttachment(attachment, false);
             });
-            
+
             /* Ignore dupe mechanic
             while(card.dupes.size() > 0 && targetLocation !== 'play area') {
                 this.removeDuplicate(card, true);
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -960,6 +960,10 @@ class Player extends Spectator {
             card.attachments.each(attachment => {
                 this.removeAttachment(attachment, false);
             });
+            
+            // Disassociate this card with all its attachmsnts
+            // They should have all been discarded and any relevant events triggered by the previous loop
+            card.attachments = _([]); 
 
             /* Ignore dupe mechanic
             while(card.dupes.size() > 0 && targetLocation !== 'play area') {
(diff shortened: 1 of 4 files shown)


server/game/player.js
```diff
@@ -960,6 +960,10 @@ class Player extends Spectator {
             card.attachments.each(attachment => {
                 this.removeAttachment(attachment, false);
             });
+            
+            // Disassociate this card with all its attachmsnts
+            // They should have all been discarded and any relevant events triggered by the previous loop
+            card.attachments = _([]);
```
alex-whitney: Are you seeing issues where `this.removeAttachment` does not remove the attachment from this card? If so, that should be fixed. This line has the potential to cause the game to enter an inconsistent state.

elstob: 👍 @alex-whitney - this is poor form and potentially destructive to the game state.

@jeremylarner, I think you need to familiarise yourself some more with the codebase and the technologies as use in general before making pull requests. While the enthusiasm is greatly encouraged, this and the other PR with no actual changes gives me concern about the quality of the code you're trying to commit.

jeremylarner: @alex-whitney

this.removeAttachment sends the attachment to the discard (or to hand if it's Ancestral), but it doesn't remove the pointer from the parent.attachment array.  I realise my submission is extremely inelegant, but as the card is leaving play and all attachments have already been removed, returning card.attachments to its initial empty state is the desired outcome, even if this is probably not the best way to do it.  If you have a suggestion for a better way to do it, I'm all ears

@elstob 

Sorry for my general ineptitude with regard to both the code and GitHub.  This is my first time using version control software, and I clearly got my branches crossed.  I'll try not to make the same mistake in future.

elstob: @jeremylarner No worries, sorry if I came across a bit harsh. The correct thing to do here is to fix up the `removeAttachment` method to ensure that it clears up this situation using the correct methods instead of brute force emptying the attachments array. This will mean things like card.owner are set to `undefined` and ensure the game state remains consistent.

It looks like there are some disabled tests in `removeattachment.spec.js` that when reenabled should provide good guidelines for how the removeAttachment method should work when complete.

alex-whitney: @jeremylarner - You may want to [squash your commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) on your PRs, as it seems like you have a bunch of irrelevant commits on each of your branches. This will keep the commit history clean and make it easier to review your changes.

PR #47, opened 2017-09-16
Ring Effects

jeremylarner:
These changes implement the effects of resolving each of the rings for a player.

Files changed:
player.js: most of the effects are in resolveRingEffects.  Air and Fire have a separate function because they require a binary choice, followed by execution of the effect chosen.


jeremylarner:
I'm going to close this, and resubmit it to try and get rid of all the irrelevant files.

PR #53, opened 2017-09-16
Play Attachment Action

jeremylarner:
There are 4 basic actions which are not card abilities but are available to the player at specific points in the game: 
- Play dynasty card from province
- Discard a duplicate to add fate to a unique character in play
- Play character from hand
- Play attachment from hand.

The current codebase implements the first of these, this is an implementation of the last one.  I've used the same format and naming as the currently implemented PlayCardAction.

The trickiest problem here is determining when it is legal to play an attachment.  We want players to only be able to play an attachment during an action window, only when they are the active player and they aren't in the middle of doing something else (like choosing targets for another ability, or resolving a reaction to another action they have taken).  My solution to this problem is slightly hacky - I ask the game what prompt is currently facing the player, and if it matches the legal time to play an attachment it returns true.  However, this requires that action windows are always labelled as such (and the tests don't like it...).  If you have a better solution, I'd love to hear it.


jeremylarner:
Update: I've changed my somewhat hacky solution into a less hacky one, although the new one relies on what might be legacy code (or at least code that seems really buggy to me).  In essence, we want to enable this action (and others like it) when there's an action window open, you're the active player, and there are no other abilities or cards being resolved.  Checking the first two are easy, as action windows create a pointer to themselves in game.currentActionWindow while they are active.  The final requirement is trickier, but all abilities add a reference to themselves (as well as whether they are currently resolving costs or effects) to game.cardAbilityStack when they start to resolve.  These are placed on top of a framework object which is placed on the stack at the start of the game (and as far as I can tell is never removed).  We want to allow this action to take place when this framework object is the only thing on the stack.

gryffon:
I'm not sure exactly what you're trying to add here. Attachments already work for the most part. When you drag a card from hand, it prompts with targets.

Are you trying to limit it to a certain phase?

jeremylarner:
The reason attachments work at the moment is due to legacy throneteki code for setup actions.  If you move out of the setup phase (play provinces and do your mulligans) then attachments don't work any more.  Lot5R doesn't have setup actions, and so we need new code which allows attachments to be played during action windows in every phase except dynasty.

PR #55, opened 2017-09-16
Added images for First Player token and Imperial Favor

jeremylarner:
Implemented First Player token image and Imperial Favor image


gryffon:
I believe the previous merges have created a conflict with this now. Please check on this and resolve it.

PR #60, opened 2017-09-18
initiate conflict

jeremylarner:
Implemented a much nicer conflict initiation prompt

### DIFF
--- a/client/GameBoard.jsx
+++ b/client/GameBoard.jsx
@@ -197,7 +197,7 @@ export class InnerGameBoard extends React.Component {
     }
 
     onRingClick(ring) {
-        this.props.sendGameMessage('flipRing', ring);
+        this.props.sendGameMessage('ringClicked', ring);
     }
 
     onFactionCardClick() {
--- a/server/game/basecard.js
+++ b/server/game/basecard.js
@@ -494,9 +494,14 @@ class BaseCard {
     getSummary(activePlayer, hideWhenFaceup) {
         let isActivePlayer = activePlayer === this.owner;
 
-        if(!isActivePlayer && (this.facedown || hideWhenFaceup)) {
+        if(!isActivePlayer && (this.facedown || hideWhenFaceup) && this.isProvince) {
             return { 
                 uuid: this.uuid,
+                facedown: true};
+        }
+
+        if(!isActivePlayer && (this.facedown || hideWhenFaceup)) {
+            return { 
                 facedown: true};
         }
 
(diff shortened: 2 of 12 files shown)


jeremylarner:
This fails tests because you choose a province by clicking on it.  Currently the game is configured not to pass any information to the client about facedown cards controlled by the other player.  This means it can't pass any information about which card has been clicked back to the game.  

I have changed that so that the game now additionally informs the client about the uuid of facedown cards, and this is enough for the game to know which province was chosen when declaring a conflict.

alex-whitney:
While no mechanics allow for it yet (that I'm aware of), revealing facedown card UUIDs would allow players to identify which facedown cards relate to cards they may have previously seen - in addition to tracking facedown cards across shuffles, swapping, etc.

Would this change allow players to identify which cards in opponents hands were drawn on specific turns, and to further track when those cards are played?

Maybe it would make more sense to select a location, rather than a province.

gryffon:
I don't think the first part could happen. There should be no correlation to a card and it's UUID, except in the same session. Tracking through shuffling would be possible though.

It *would* allow people to know when cards are drawn and track them. Though that's a thing in other popular online games. There is a program that will do just that in Hearthstone, but that just boils down to a memory helper.

Selecting a location would be more optimal. We know where locations are when we are dragging the cards to the provinces in the first place, so there has to be some precedence for know what location you're over, and that would conveniently work even when there is no card there(solo play).

jeremylarner:
How easy is it for players to actually get access to uuid information, even if the client has access to it?  Can you hack your own React code in instead of using the server's?

Clicking locations might work.  The problem is that the location is entirely covered by the cards in it.  Any click in that area is picked up by the clickOnCard code.

One solution would be for the game to pass a dummy uuid (representing the location of that card) to the client, and then intercept that dummy code when it's picked up by the initiate conflict code. It's a bit hacky (because you're passing around fake uuids), but it would mean that there's no way the player can glean any useful info from the uuid.

jeremylarner:
In terms of tracking draws etc., I'm pretty sure I can make it so the only uuids which are available are for Province cards.  There's no shuffling of those at the moment, although Alex is right that it could be introduced in the future.

jeremylarner:
I changed it so the uuid is only passed to the client for provinces.  It even passes the tests now! ;)

gryffon:
I think this may be a reasonable compromise. We'll definitely want to look in testing to see if this is providing more information than intended, however.

jeremylarner:
I thought of an alternative solution:

Instead of the game passing the uuid of facedown cards to the client, it passes their location.  This shouldn't give any additional information to nefarious players, as they know the location of whatever they clicked.

It means changing the client side clicking function to pass uuid normally, or location when uuid is unknown (actually, passing both is probably easier and less likely to cause bugs), and then changing the click handler in Game and getting it to intercept locations and substitute the appropriate uuid.

Thoughts?

gryffon:
I've spun off your alternative solution to an issue that we can come back to later.

PR #61, opened 2017-09-19
Role Card Implementation

jeremylarner:
This updates the deck editor and game lobby code to support role cards.  Any decks built from now on recognise role cards and will pass them through Game to Player, which will recognise them and pass info to the GUI.

I haven't got the GUI to work with them yet.  Also, note that it won't work retroactively on decks previously built, as role cards need to be stored as a separate attribute in the deck object.


jeremylarner:
Rolecards will now also appear on the gameboard where present

PR #69, opened 2017-09-20
Play character action

jeremylarner:
Allows user to play characters from hand during action windows, and uses a menu to find out how much additional fate they wish to place on that character.  Also checks whether they wish to play the character at home or into the conflict during conflicts.


gryffon:
This has come up with a conflict due to prior merges, please resolve.

jeremylarner:
done

PR #95, opened 2017-09-21
manual mode

jeremylarner:
Added a manual mode, enabled through the /manual chat command, which adds a button to action windows to allow players to use unimplemented cards and manually resolve conflicts.


gryffon:
Could you add this to the documentation on the About page as well?

client/About.jsx

jeremylarner:
I've added some documentation

PR #154, opened 2017-09-25
Superfluous message removed

jeremylarner:
Currently, players get a message about 0 bids, which shouldn't be displayed. This fixes that.


jeremylarner:
Also added a /claim-ring fix for #155

PR #169, opened 2017-09-27
implementing cards documentation

jeremylarner:
Updated with L5R examples and terminology


gryffon:
Did we want to arrange cards by pack, instead of type?

gryffon:
Yeah, I think learning from our peers and doing it right the first time is probably the way to go. I think we're confidant in wanting to do it that way, so let's start that way instead of one way and then changing it.

jeremylarner:
This has been updated and is ready to go now, along with the backport PR

PR #170, opened 2017-09-27
Update effects 

jeremylarner:
Changed effects.js to match L5R attributes, and added some effects we will need for implementing cards.

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -107,8 +107,8 @@ class DrawCard extends BaseCard {
 
     modifyGlory(amount, applying = true) {
         /**
-         * Modify the military skill.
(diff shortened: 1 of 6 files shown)


gryffon:
I've re-run the pull request build, and it's now actual failures.

jeremylarner:
This passes now

server/game/drawcard.js
```diff
@@ -106,6 +105,20 @@ class DrawCard extends BaseCard {
         }
     }
 
+    modifyGlory(amount, applying = true) {
+        /**
+         * Modify the military skill.
```
gryffon: Comment looks it needs to be updated.

jeremylarner:
Thanks for spotting that.  I think it's good to go now.

PR #196, opened 2017-09-29
Vengeful Oathkeeper

jeremylarner:


gryffon:
I'm a little fuzzy, but reactions are still optional in the code, right? Meaning, just because the situation has been made, doesn't mean the card is going to jump into play, correct?

jeremylarner:
Right, the game has two kinds of reaction: reactions and forced reactions.  This is a reaction, so the game will prompt the player whenever it triggers, and they can choose to use the ability or pass.

gryffon:
Thanks, just double-checking

PR #215, opened 2017-10-01
Yasuki should trigger from opponent's holdings

jeremylarner:


gryffon:
Should we consider making the holding check a more general function? There's at least a couple of cards that look for that.

jeremylarner:
You mean put it in Game and returning the number of holdings in play for both players?  I thought about it, but there hida guardian needs a function for just that player's holdings, so we need the Player function in any case.

We could do both, but I don't know that it's worth it.  Happy to defer to your judgement though, if you'd prefer it implemented that way.

jeremylarner:
Had a report that putting the card to the bottom isn't working properly, so I've implemented a workaround.  We probably need to look at implementing throneteki's moveCard synchronicity fixes in the future though.

gryffon:
Well, more of a function to check an individual player for holdings. Hida Guardian could call it just for the player, and Yasuki and Shuichi can call it for both.

jeremylarner:
So I did implement exactly that when I originally implemented the card.  The card doesn't have code for counting holdings, it uses a general function in Player

PR #291, opened 2017-10-07
Ancestral Lands

jeremylarner:


jeremylarner:
This needs the province strength code in one of the other PRs.  If you could rerun the tests after you've approved that, I think it should pass

PR #304, opened 2017-10-07
cards which leave play shouldn't be participating

jeremylarner:


jeremylarner:
This also needs some code from another PR

gryffon:
Looks like this has a lint error now

jeremylarner:
fixed it

PR #311, opened 2017-10-08
Control changes

jeremylarner:
- Backported Throneteki code which moves allCards to Game instead Player
- Refactored take control code for characters and attachments
- Disabled the put into play tests.  I can't find a nice way to duplicate these tests under the new format without rewriting the code to pass the tests, which feels like the tail wagging the dog.  Jasmine doesn't deal with arrays very well, and _arrays are even worse...


gryffon:
Looks like this has some conflicts, probably from other merges

jeremylarner:
conflicts resolved

PR #340, opened 2017-10-09
added clickToActivate to monks

jeremylarner:


jeremylarner:
Don't need this because of #354

PR #444, opened 2017-10-13
Update target API to allow opponent to target

jeremylarner:
Also updated Master of the Spear to use the new API


gryffon:
Should Way of the Crab also use this?

PR #463, opened 2017-10-14
Support rings with multiple elements

jeremylarner:
This change implements rings with multiple elements and gives players a choice about which ring effect to resolve.

It also introduces support for resolving multiple elements simultaneously, with first player choosing order in those circumstances


gryffon:
Looks like this has a conflict

jeremylarner:
Conflict has been resolved

Issue #368, opened 2017-10-11
font issue with political and military


jeremylarner:
'political' and 'military' should be interpreted and replaced with the special font symbol like rings are, but they are rendering as blank atm...

jeremylarner:
Fixed by #496

PR #496, opened 2017-10-16
make cards linkable from chat

jeremylarner:
Also a fix for #368


gryffon:
Is the change to the css to fix something? There are some places that use that already and I don't want those to break.

jeremylarner:
I was just changing them to match what's in the game chat, but I can change the game chat to match the css instead

Issue #598, opened 2017-10-20
Ambush


vmironovich:
Ambush allows to put face down cards from provinces into play.

jeremylarner:
Thanks for the report - fix submitted #603

PR #603, opened 2017-10-20
bugfix for ambush and cavalry reserves

jeremylarner:
fix for #598


vmironovich:
Cavalry reserves doesn't require the bugfix, as at the moment no cards in dynasty discard can be facedown.

jeremylarner:
That's true.  I templated the two abilities very similarly, so in my mind the bug fix should apply to both of them.  It won't do any harm.

Issue #445, opened 2017-10-13
You can initiate Master of Spear's ability when there are no legal targets


jeremylarner:
Ability Context is set by AbilityResolver, but the game checks to see if there are legal targets in Player.playCard first.
However, some cards need to know the source of an effect in order to determine whether they are can be targeted by that effect. At the moment, they get that information from Ability Context, but that means they are regarded as legal targets for the pre-initiation check.
The game correctly stops you from targeting them when you get to targets.

jeremylarner:
Fixed in #623

PR #623, opened 2017-10-21
Add checks for gameAction immunity to pre-action checks

jeremylarner:
Fix for #445


gryffon:
Do we have a fix for this request?

PR #678, opened 2017-10-25
Pre-targeting on abilities

jeremylarner:
Changed the ability resolver to do targeting before choosing/paying costs.  This allows:
 - implementation of cards which require targets to be selected to reduce their costs (Daimyo's Favor)
 - players to cancel abilities which they've clicked on once they realise that the target they wanted can't be chosen, or just change their mind

In some cases, payment of costs may impact the targets available, so where that is the case, players can choose to pay costs first.  If they target first, and then payment of costs invalidates some or all of those targets, they will be asked to rechoose any invalid targets.


gryffon:
Looks like this has test errors

PR #756, opened 2017-11-01
 discard characters with no fate bug

jeremylarner:


jeremylarner:
Not sure why this is failing tests - it passes locally

PR #792, opened 2017-11-02
Staunch Hida implementation

cody1024d:


jeremylarner:
This looks fine to me - but needs to go into the dev branch, not master?

cody1024d:
I agree.  Does need to go into Develop.  I've made the change @jeremylarner  @gryffon

PR #800, opened 2017-11-03
Pass source to bow, (dis)honor, and ready effects

jeremylarner:


gryffon:
Needs to updated, due to a conflict

PR #801, opened 2017-11-03
change allowGameAction to not use stack

jeremylarner:


gryffon:
Has some test failures

Issue #776, opened 2017-11-02
Shosuro Miyako triggers on herself


Arbuzio:

Arbuzio:
and also she dont react on cards played from hand I think...

jeremylarner:
Fixed in #810

PR #810, opened 2017-11-03
Update ability resolver to use AbilityContext and fix onCardPlayed bug

jeremylarner:
#Fix for #776


jeremylarner:
I'm just looking at Sentry, and there's a bug that I think will get worse, not better with these changes - gonna try and fix it

jeremylarner:
Ok, hopefully this fixes the Yojimbo bug as well now

PR #817, opened 2017-11-03
Card/goblin sneak

cody1024d:


cody1024d:
@gryffon  @jeremylarner  I think this follows the proper folder structure that ya'll wanted?

PR #823, opened 2017-11-04
Fix moveToConflict bug

jeremylarner:


jeremylarner:
That's what the current code does, which is a problem.  I changed it (I think!?!) so that being bowed stops characters being declared, but not participating.

PR #841, opened 2017-11-05
Card/tears of amaterasu

cody1024d:


cody1024d:
@jeremylarner Why a condition on number of attackers?  Technically, the ability does still trigger, even if there are 0 characters, it just has no effect?

jeremylarner:
No, technically, an ability can't be triggered if it wouldn't affect the game state. So if an attack were somehow declared with 0 attackers, then you wouldn't be allowed to trigger the ability. Night Raid (for example) has a condition on the opponent's hand size for exactly this reason.

cody1024d:
Ah, fair point.  Ok, I will definitely add the check in then!

PR #853, opened 2017-11-06
Refactor context.targets for AbilityResolver

jeremylarner:
The context object for AbilityResolver has an issue in that most code expects targets to be BaseCard objects.  However, in L5R some cards require a 'select' choice or choosing a ring during targeting.  At the moment, they are all stored in context.targets, but this causes bugs and a lot of ugly code.

Solution: add context.rings and context.selects to store rings and select choices respectively.  I've also created a context.ring and context.select which serve the same purpose that context.target does for cards.

Still to do: This breaks the prompt controls for cancel abilities.  Cards which target rings or have 'select' choices will no longer display those choices.

### DIFF
--- a/server/game/AbilityTargets/AbilityTargetCard.js
+++ b/server/game/AbilityTargets/AbilityTargetCard.js
@@ -15,7 +15,7 @@ class AbilityTargetCard {
 
     resolve(context, pretarget = false, noCostsFirstButton = false) {
         let otherProperties = _.omit(this.properties, 'cardCondition');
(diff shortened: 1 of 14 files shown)


gryffon:
Should this wait on the TODO?

jeremylarner:
The reason I haven't done the TODO is that the code that I need to change to fix it has been changed in the master branch.  So I'd rather wait until we merge this into master, and then I'll make the changes.

If I change it now, it's just going to create lots of messy conflicts to resolve when you merge them.

PR #855, opened 2017-11-06
Implementation for Disdainful Remark

cody1024d:


jeremylarner:
Also, if you could make sure you put the card name in the PR title, that really helps for keeping track of things.

cody1024d:
Ah you're totally right!  I keep forgetting about the whole, needs to affect board-state thing.

PR #900, opened 2017-11-08
Update to react 16

gryffon:


gryffon:
Most of them were already updated previously.

PR #906, opened 2017-11-09
Card/smoke and mirrors

CDMayberry:
Works now. One thing, the button 'Pay Cost First' appears during the targeting, but it isn't necessary as this has no cost associated. However I didn't see any settings to disable it. Is there a way? If not then this should be good to go.


jeremylarner:
Because of the changes I made to this, it won't work without #911

PR #966, opened 2017-11-12
Stoic Magistrate

cody1024d:


cody1024d:
*Note* This hasn't been tested yet, because cards are not on fiverings.

jeremylarner:
You can test cards not in the api by changing the id to match a different card (e.g. in this case you could use 'swift-magistrate') The base stats of the card will be wrong, but you should be able to see if the ability is doing the right thing.

cody1024d:
Not yet.  Haven't had the time :/

PR #1024, opened 2017-11-15
Implementing frostbitten crossing

cody1024d:


cody1024d:
@jeremylarner   I think either of those solutions is ok.  I do like the idea of having a helper function to remove all attachments, if only because that may become a more-common use case as the game evolves.  I'll look into how I would use raiseMultipleEvents to fix this (hopefully) tonight.

PR #1088, opened 2017-11-17
UI: make unselectable rings grey out during select prompts

jeremylarner:
image
image


jeremylarner:
I'm also considering greying out all rings bar the contested one during conflicts, but I'm not sure how easy that is to do with the current infrastructure (it seems to be tied quite tightly to the prompt state).

Any thoughts?

jeremylarner:
image

jeremylarner:
The amount of fate on rings, and which ones are available is really important in making decisions throughout the turn.  I'm not sure that hiding that info in a popup box is a good idea.  I could see it with chat, although I know that players rely on that to help them keep track of the game state also.

Issue #1085, opened 2017-11-17
Censure


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/32954925-682ff67e-cbac-11e7-8b47-b2d24af1b34e.png)

CDMayberry:
Working on this one

PR #1125, opened 2017-11-21
Censure

CDMayberry:
Implements #1085


CDMayberry:
Used Voice of Honor as the basis, wasn't 100% sure on it either.

PR #1133, opened 2017-11-22
Implementing Hiruma Skirmisher

### LINKED ISSUE #1077: Hiruma Skirmisher
image

SnarkAttack:
Closes #1077


jeremylarner:
It occurs to me that you may not have tested it because api data for this card isn't available from 5rdb.  I'd recommend changing the id to match another card that is in the database, and testing it that way.

SnarkAttack:
I tested it last night, the data for Hiruma Skirmisher is available

jeremylarner:
And it worked, with that code?  Because I can't see how that effect will grant Skirmisher (or anyone else) Covert...

SnarkAttack:
So it did work even with the match, since the only target was itself (you just picked it). But you're right, not having the match is probably the cleaner way. Updated in the most recent commit.

SnarkAttack:
Updated again. I apologize for not catching this when I tested.

PR #1136, opened 2017-11-22
Implemented Public Forum

### LINKED ISSUE #1069: Public Forum
image

SnarkAttack:
Closes #1069 

I added a hasHonorStatus field to province cards that is only used in the interrupt for Public Forum. It might be smart to at some point turn it into a list of statuses if it seems like this becomes a more popular design space for them.


jeremylarner:
This probably requires #1142 before it can be merged

PR #1189, opened 2017-11-24
Add honor token support

### LINKED ISSUE #1142: Token support for Dynasty cards needed
In order to support implementation of Public Forum, we need client-side changes which look for the tokens object (which is currently passed to the client for all cards), and render an honor token when tokens.honor is 1.

I'm not sure how this will fit with the current CardCounter class (and the token classes it uses).  It may be worth looking at integrating Fate tokens, honored/dishonored tokens and honor tokens to all use the same extendable system, in case we get additional tokens in the future.

gryffon:
Closes #1142


client/GameComponents/Card.jsx
```diff
@@ -132,6 +133,7 @@ class Card extends React.Component {
         var counters = {};
 
         counters['card-fate'] = card.fate ? { count: card.fate, fade: card.type === 'attachment', shortName: 'F' } : undefined;
+        counters['card-honor'] = card.honor ? { count: card.honor, fade: card.type === 'attachment', shortName: 'F' } : undefined;
```
jeremylarner: I don't think card.honor is something which is used by either the client or the server, but I'm assuming this is here for future-proofing!  I think if we ever needed to put honor on anything else, card.tokens is probably the best way to do it.

PR #1196, opened 2017-11-24
Conflict hand

dastardlychimp:


jeremylarner:
I played around with this a bit.  I really like the way the hand animation works, and I also really like the icons you've made.  

However, I think the changes you've made to the way the UI works probably need some revisions.  I feel like when playing the game, the player wants to be able to refer to cards in their hand frequently, and needing to click back and forward between the hand and the general play area might make the interface feel unwieldy.  As you've identified, there are also issues when trying to select targets in your hand and on provinces simultaneously.

Could the stuff you've done be adapted to follow the interface systems we have at the moment more closely.  Could we make the hand you've created smaller, and have it draggable much like the hand we've got at the moment, just much prettier?  I'd really like to have the player able to see what's in their hand at all times.

As I said above, I really like the work you've done on icons, and would like to think a bit more on how it could be integrated into our current UI.  Could we use them in the current player prompt?  Mostly there we just need a cancel and a done button.  I also made use of a 'pay costs first' button, but I don't know what a more symbolic representation of that might look like...

dastardlychimp:
Now, I totally agree that switching between the hand and the board is a pain. However, I'm not sure how to make a smaller interface pretty. Part of the reason I thought of having the hand open and close, was to clear up room for the game board, so the cards there could be made larger or other revisions.  We could bind a hot key ('h') to opening and closing the hand, but you're right; it's probably not enough. I might not be understanding your idea correctly though.

As for the icons... I don't know if it would work with the current prompt. I mean, they are buttons, so you could replace buttons with these icons, but there are a bunch of prompts that there are no icons  for and an icon can't explain -- such as when it prompts you to choose a card and it gives you five card names. I'm not sure switching between the green block buttons and the circle icons would work. Perhaps, the cancel and done icons could be placed in the bottom of the prompt and other options show above that.

jeremylarner:
No, I think you've understood my suggestion correctly.  When you say we can't implement what you've done on a smaller scale, is that for technical reasons (i.e. it's impossible to create an animated hand with stylised cards at the scale that we currently have the conflict hand), or are you saying you don't think it would be a good change?

With regard to the circle icon changes, you're right and my idea is probably not well thought through.  Card prompts I do plan to replace though (I'd rather show the cards the player is choosing between than have them on buttons in a menu), but we use button menus for lots of other things, and we'd need to do a review and see if we could replace them adequately with circle icons.

dastardlychimp:
It could be implemented on a smaller scale, but I don't really see the point of it on a smaller scale. The animation is nice and all, but really the idea was to make the cards bigger (readable) and give more room for the game board. Animation on the cards doesn't seem great if you have to look at the top right corner to read the card's text.

I agree with you here. Some of the menu buttons will be shifted into other components. In the future, people will choose the actual cards, and maybe a dial component for choosing the amount of fate to place on a character, etc. Though, like you mentioned, it would be difficult to find icons for 'pay costs first', at least icons that are understandable to people new to the game.
But the circle icon is created, and it won't be difficult to use when we want to, if we choose to do so.

gryffon:
Thanks for your suggestions and works, we've done a UI overhaul and it supercedes this PR.

Issue #1163, opened 2017-11-23
Backhanded Compliment


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/33183211-d5565966-d06e-11e7-8390-bb2ae818d02d.png)

CDMayberry:
Working on this.

PR #1210, opened 2017-11-25
Backhanded Compliment

CDMayberry:
I got this working, though I see a lot of places uses promptWithHandlerMenu rather than target -> select. Would it be better to re-write it to use that instead?

Implements #1163

### DIFF
--- a/server/game/cards/02.4-TCT/BackhandedCompliment.js
+++ b/server/game/cards/02.4-TCT/BackhandedCompliment.js
@@ -3,16 +3,12 @@ const DrawCard = require('../../drawcard.js');
 
 class BackhandedCompliment extends DrawCard {
     setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
-        let choices = {};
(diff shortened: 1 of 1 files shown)


jeremylarner:
I'm pretty sure this works but it needs test fixes, and I'm not sure it's the best way to do it from a UI point-of-view.  I think a better approach might be to make the select prompt between 'Me' and 'My Opponent'.

If you want to go with player names, you need to add 'getPlayers' to the jasmine game object in the failing test.  If you're not sure how to do that, I can submit a PR to your branch for it.

CDMayberry:
I agree from a UI perspective it would be best to do Me/My Opponent, though I did like that this worked with N players. I'll go ahead and swap to the Me/My Opponent, I'll rewrite the test if it's still needed after I rework this.

CDMayberry:
No, your completely right, I just forgot how choices worked.

CDMayberry:
Updated and tested.

PR #1223, opened 2017-11-27
Kaiu Inventor

jeremylarner:


jeremylarner:
So I found a bug with this card where the lasting effect from inventor remains on the card, even after it has been discarded.  I've tried to fix this by making changes to the effects engine, but I don't really understand how it works. So please look at those changes particularly and see if you can find anything that might cause problems.

PR #1317, opened 2017-12-05
Master of Gisei Toshi

### LINKED ISSUE #1250: Master of Gisei Toshi
image

CDMayberry:
Closes #1250


jeremylarner:
I've changed the way allowGameAction works in another PR (#1297), so the cannotTriggerAbilities effect won't work the way you've coded it if/when that PR is merged.  I'll wait until that one is merged before I fix this card though.

Apart from that it looks fine.

PR #1339, opened 2017-12-07
Crisis Breaker

### LINKED ISSUE #1245: Crisis Breaker
image

SnarkAttack:
Closes #1245

I had to make a few changes to conflict.js, mostly due to Crisis Breaker fighting at Massing At Twilight, although I'm guessing this won't be the last time something like this is relevant. Noted changes are making a copy of calculateSkillFor where the skillFunction added in #1300 will never change and splitting attackerSkillModifier into attackerMilitarySkillModifier and attackerPoliticalSkillModifier (did same for defender). 

Added two functions, compareMilitary() and comparePolitical(). They return the difference between attacker and defender skill of the given conflict type ( < 0 => attacker is lower, = means equal, > 0 means attacker higher).

### DIFF
--- a/server/game/cards/02.5-FHNS/CrisisBreaker.js
+++ b/server/game/cards/02.5-FHNS/CrisisBreaker.js
@@ -28,7 +28,7 @@ class CrisisBreaker extends DrawCard {
             target: {
                 activePromptTitle: 'Choose a character',
                 cardType: 'character',
-                cardCondition: card => card.location === 'play area' && card.hasTrait('berserker') && (!card.isParticipating() || card.bowed)
+                cardCondition: card => card.location === 'play area' && card.hasTrait('berserker') && (card.allowGameAction('moveToConflict') || card.allowGameAction('ready'))
             },
             handler: context => {
                 this.game.addMessage('{0} uses {1} to ready {2} and move it into the conflict', this.controller, this, context.target);
--- a/server/game/cards/02.5-FHNS/CrisisBreaker.js
+++ b/server/game/cards/02.5-FHNS/CrisisBreaker.js
@@ -7,7 +7,7 @@ class CrisisBreaker extends DrawCard {
             condition: () => {
                 const currentConflict = this.game.currentConflict;
 
-                if (!currentConflict || currentConflict.conflictType !== 'military') {
+                if(!currentConflict || currentConflict.conflictType !== 'military') {
                     return false;
                 }
 
--- a/server/game/cards/02.5-FHNS/CrisisBreaker.js
+++ b/server/game/cards/02.5-FHNS/CrisisBreaker.js
@@ -1,10 +1,44 @@
 const DrawCard = require('../../drawcard.js');
 
 class CrisisBreaker extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities() {
+        this.action({
+            title: 'Ready and bring into play',
+            condition: () => {
+                const currentConflict = this.game.currentConflict;
+
+                if (!currentConflict || currentConflict.conflictType !== 'military') {
+                    return false;
+                }
+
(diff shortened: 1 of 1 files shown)


jeremylarner:
Would it not be easier to just change the SkillFunction being used to getMilitarySkill() in the condition of the card ability?

SnarkAttack:
Good idea, but actually implemented the change to skillFunction inside of calculateMilitary() and calculatePolitical()

server/game/cards/02.5-FHNS/CrisisBreaker.js
```diff
@@ -1,10 +1,26 @@
 const DrawCard = require('../../drawcard.js');
 
 class CrisisBreaker extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities() {
+        this.action({
+            title: 'Ready and bring into play',
+            condition: () => this.game.currentConflict && this.game.currentConflict.conflictType === 'military' &&
```
devofdisaster: This seems unnecessarily complicated - wouldn't something similar to this be simpler? It would only recalculate the conflict skill twice, as opposed to four times, allowing us to get rid of the comparison methods completely:

    () => {
        const currentConflict = this.game.currentConflict;
    
        if (!currentConflict || currentConflict.conflictType !== 'military') {
            return false;
        }
    
        const originalSkillFunction = currentConflict.skillFunction;
    
        currentConflict.skillFunction = (card) => card.getSkill('military');
        currentConflict.calculateSkill();
    
        const conditionFulfilled = currentConflict.attackingPlayer === this.controller ?
            currentConflict.attackerSkill < currentConflict.defenderSkill :
            currentConflict.defenderSkill < currentConflict.attackerSkill;
    
        currentConflict.skillFunction = originalSkillFunction;
        currentConflict.calculateSkill();
    
        return conditionFulfilled;
    }

server/game/effects.js
```diff
@@ -441,6 +441,21 @@ const Effects = {
             }
         };
     },
+    changeConflictSkillFunction: function(func) {
+        return {
+            apply: function (card, context) {
+                if(context.game.currentConflict) {
+                    context.skillFunction = context.game.currentConflict.skillFunction;
```
devofdisaster: Minor nitpick, but I'd change the name of `context.skillFunction` to something like `originalSkillFunction`, to make it easy to recognize what it's used for, in case someone later stumbles upon it.

jeremylarner: I think this is a good idea, and I'm going to change it in my Massing at Twilight implementation.

server/game/conflict.js
```diff
@@ -20,11 +20,15 @@ class Conflict {
         this.elements = [];
         this.attackers = [];
         this.attackerSkill = 0;
-        this.attackerSkillModifier = 0;
+        this.attackerMilitarySkillModifier = 0;
+        this.attackerPoliticalSkillModifier = 0;
```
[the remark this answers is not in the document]
SnarkAttack: I 100% agree, but I didn't want to rip them out since that wasn't specifically my task here and I didn't know if there was going to be a use case one day

server/game/cards/02.5-FHNS/CrisisBreaker.js
SnarkAttack: It's definitely cleaner for one card, and if we want to go this way we can. I wrote the other methods under the assumption that this would not be the only card to want to compare strength in a case where that might not be the conflict comparison, and having the functions implemented in conflict prevents us having to do this every time. Again, not opposed to this plan, just trying to future proof. At the very least I could just make the results of compareMilitary it's own value so we don't have to calculate twice

SnarkAttack:
Blew everything up and just moved skill calculations into cards. Thank you devofdisaster for making that job a lot quicker!

PR #1403, opened 2017-12-12
Remove 'Done' from a lot of SelectCardPrompts

### LINKED ISSUE #1284: Endless plains trigger didn't enforce removal of character on the field
Turn 1. Playing unicorn again. Opponent is playing crab. I have a Utaku Mediator and Utaku Yumino. Opponent controls Kaiu Envoy and a Stoic Magistrate. Opponent attacks into an Endless plains with the kaiu envoy. It is a void military conflict. This is the first attack this turn. Endless plains triggers and is activated by me, breaking the province and discarding altansarnai who was placed over it. But from the looks of it the opponent did not receive a prompt to discard his kaiu. Or if he did it did not resolve correctly.

jeremylarner:
- Instead, a Cancel Prompt button will be displayed while in manual mode.
- Done button is still used for multiselect prompts, but won't complete the prompt if pressed when no cards are selected
- Cost selections should have a Cancel button

Fixes #1284


jeremylarner:
Needs a fix to make it compatible with Five Fires

PR #1423, opened 2017-12-13
Implemented Kuni Ritsuko

SnarkAttack:
Closes #1343

PR #1432, opened 2017-12-15
Meditations on the Tao Update

SnarkAttack:
Made the same changes requested in #1423


jeremylarner:
Actually, now I look at this again, I realise I screwed up.  Card.modifyFate calls the correct raiseEvent, so this change isn't needed.  Your previous implementation of Kuni Ritsuko was also fine.  I'm really sorry!

PR #1442, opened 2017-12-15
add putSelfIntoPlay cost

jeremylarner:


jeremylarner:
This needs changing to work with the new costs code

PR #1464, opened 2017-12-18
Implemented Fushicho

### LINKED ISSUE #1439: Fushicho
image

artemmikhalitsin:
Closes #1439
-I made an assumption as to what the pack folder will be called.
-Is there a better way to have the character enter play with 1 fate on it already?


jeremylarner:
It's not clear how the card is supposed to work, and whether it triggers abilities which reference placing fate on a character.  I've changed your code so that it just sets the amount of fate to 1, without creating any events associated with that.  Also, could you change the directory to 03-DotV? It's not part of the imperial cycle, so I think it deserves it's own number.

PR #1476, opened 2017-12-19
Kaito Kosori

jeremylarner:


jeremylarner:
The effects.js change will need rewriting to work with #1731 
I'd also be much happier if this had a test

PR #1523, opened 2017-12-29
Feature/emoji in chat

zacharyp:
Normal emoji added in chat via emoji-js library, https://github.com/iamcal/js-emoji

chat_emoji


jeremylarner:
This would probably be more used in lobby chat if you could add it there also.

zacharyp:
aah, ok.  Will do.  I also noticed I mistakenly branched off of master instead of develop.  I reopen with changes

zacharyp:
new PR with lobby chat as well: https://github.com/gryffon/ringteki/pull/1524

PR #1569, opened 2018-01-15
Ambush bugfix

### LINKED ISSUE #1565: Ambush lets you select 3 characters costing up to 6
More generally, the MaxStatCardSelector doesn't check against the numCards parameter

artemmikhalitsin:
Closes #1565 

Some notes:
Possibly, MaxStatCardSelector could inherit from UpToXCardSelector, or reworked to have a constructor that looks similar to other selectors that select some # of cards


artemmikhalitsin:
Yep my bad. I left your commit so that initiateConflict functions properly, its ok?

PR #1574, opened 2018-01-16
Katana of Fire implementation. 

### LINKED ISSUE #1438: Katana of Fire
image

artemmikhalitsin:
Requires #1573 for tests to pass.
Closes #1438


jeremylarner:
This will need updating to deal with ring effects introduced in #1684

jeremylarner:
Also need to check that Restricted is parsed correctly

PR #1581, opened 2018-01-16
Honoring already honored characters bugfix

### LINKED ISSUE #1578: Savvy Politician + Pride
If you have an honored Savvy Politician and give it pride with for example Magnificent Kimono, and the Savvy Politician is honored, you can still fire it's ability upon winning a conflict. 

This is one of the few edge cases where the fact that Savvy Politician is honored prevents it from triggering, without there being any player targeting involved. (cards like way of crane)

artemmikhalitsin:
Requires #1579 and #1580 for test to pass
Closes #1578

### DIFF
--- a/server/game/drawcard.js
+++ b/server/game/drawcard.js
@@ -403,6 +403,9 @@ class DrawCard extends BaseCard {
     }
 
     honor() {
+        if(!this.allowGameAction('honor')) {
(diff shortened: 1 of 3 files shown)


jeremylarner:
This has been incorporated into #1488

PR #1582, opened 2018-01-16
card list/ring list toInclude comparator

artemmikhalitsin:
Can be used to check if a card or a ring is in a list


artemmikhalitsin:
Apparently toContain() already exists

PR #1588, opened 2018-01-17
Repurpose Game.applyGameAction()

jeremylarner:
This PR repurposes applyGameAction to implement common effects in the game in a more elegant way.  It allows multiple effects to be combined easily, and moves a lot of similar code (e.g. parameters for events) to somewhere it can be changed once rather than having to search through the entire codebase.

It deprecates a number of functions in Player and Conflict, and should make those objects easier to navigate.  It also centralised all the allowGameAction checks to Events, so those shouldn't be necessary in implementing cards.


jeremylarner:
Needs #1488 and #1586 to pass tests

PR #1590, opened 2018-01-17
Remove result from Event

jeremylarner:
A long standing problem with the game engine is that there is no way to track whether an effect has resolved successfully, or a cost has been successfully paid.  I partially implemented a fix for this where event handlers returned a result object so that the game could verify that the handler had resolved, but I don't think this is the best way to solve this particular issue, so this PR removes some ground work I did for it.

I also used this system for thenEvents, so I've refactored thoses and rewritten Bayushi Kachiko.  Test for Bayushi Kachiko also included


jeremylarner:
Needs #1569 to pass tests

jeremylarner:
@gryffon can you requeue this for tests and see if it passes?

PR #1601, opened 2018-01-31
Clear out stale games that have not started within an hour

zacharyp:
Took the code from Throneteki (https://github.com/cryogen/throneteki/pull/1593), but changed the stale timeout time to an hour instead of 15 minutes.  I know there are times in the early morning where I wait more than 15 minutes for someone to join my game ;)


jeremylarner:
I'm not really very familiar with the lobby code, so I don't feel comfortable reviewing this, but it looks like your check for empty games looks for games with no players which have been going for more than 24 hours - is that right? 24 hours seems a long time to keep open empty games...

zacharyp:
I'm happy to reduce the time from 24 hours down to an hour for both checks.

PR #1803, opened 2018-04-06
tests added for Karada district (issue in favored mount + born in war…

Syrail:
… test)


jeremylarner:
needs #1804 to pass tests

PR #1903, opened 2018-06-15
Indicators to show whether a character has abilities, and whether they have been used

jeremylarner:
image


jeremylarner:
I need to work out how to change the default location of the indicator on attachments as at the moment it isn't visible

gryffon:
Closing to reevaluate based on new UI

PR #1904, opened 2018-06-15
Added Kudaka

### LINKED ISSUE #1896: Kudaka
image

CDMayberry:
resolves #1896

### DIFF
--- a/server/game/cards/04.1-BotK/Kudaka.js
+++ b/server/game/cards/04.1-BotK/Kudaka.js
@@ -1,7 +1,7 @@
 const DrawCard = require('../../drawcard.js');
 
 class Kudaka extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities(ability) {
         this.reaction({
             title: 'Gain 1 fate and draw 1 card',
             limit: ability.limit.perRound(2),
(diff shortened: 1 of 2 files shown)


server/game/cards/04.1-BotK/Kudaka.js
```diff
@@ -1,10 +1,19 @@
...
 
 class Kudaka extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities(ability) {
+        this.reaction({
+            title: 'Gain 1 fate and draw 1 card',
+            limit: ability.limit.perRound(2),
+            effect: 'gain 1 fate and draw 1 card',
+            when: {
+                onClaimRing: (event, context) => event.conflict && event.conflict.ring.element === 'air' && event.player === context.player
```
jeremylarner: This will only explicitly check for the air ring - you need to check for any additional elements.
You should use event.conflict.hasElement('air'), I think

CDMayberry: Ah, i forgot that it counts any additional elements as being claimed. I'll add that and some tests tomorrow.

CDMayberry:
Added the hasElement and a basic test. Doesn't include a test scenario with seeker of knowledge atm, I'll work on that next, but functionally wise it should be good to go.

jeremylarner:
Test is broken :P
Card implementation looks fine though

Issue #1999, opened 2018-06-20
Soul Beyond Reproach


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/41689526-eb8e4a3a-74e8-11e8-91ae-2aad9613091d.png)

PR #2062, opened 2018-06-25
TEaF: Soul Beyond Reproach

CodeBaboon:
Implements #1999

### DIFF
--- a/server/game/cards/04.4-TEaF/SoulBeyondReproach.js
+++ b/server/game/cards/04.4-TEaF/SoulBeyondReproach.js
@@ -1,10 +1,22 @@
 const DrawCard = require('../../drawcard.js');
 
 class SoulBeyondReproach extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Honor a character, then honor it again',
+            target: {
+                cardType: 'character',
+                controller: 'self',
+                gameAction: [
+                    ability.actions.honor(),
+                    ability.actions.honor()
+                ]
+            },
+            effect: 'honor {0}, then honor {0} again'
+        });
     }
 }
 
-SoulBeyondReproach.id = 'soul-beyond-reproach'; // This is a guess at what the id might be - please check it!!!
+SoulBeyondReproach.id = 'soul-beyond-reproach';
 
 module.exports = SoulBeyondReproach;
--- a/test/server/cards/04.4-TEaF/SoulBeyondReproach.spec.js
+++ b/test/server/cards/04.4-TEaF/SoulBeyondReproach.spec.js
@@ -1,0 +1,70 @@
+describe('Soul Beyond Reproach', function() {
+    integration(function() {
+        describe('Soul Beyond Reproach\'s ability', function() {
+            beforeEach(function() {
+                this.setupTest({
+                    phase: 'conflict',
+                    player1: {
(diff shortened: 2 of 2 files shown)


server/game/cards/04.4-TEaF/SoulBeyondReproach.js
```diff
@@ -1,10 +1,22 @@
...
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Honor a character, then honor it again',
+            target: {
+                cardType: 'character',
+                controller: 'self',
+                gameAction: [
+                    ability.actions.honor(),
+                    ability.actions.honor()
```
jeremylarner: The way you've done this, it will apply two simultaneous honor effects to the target, whereas the card says 'Then, ...' so we really want them to be sequential.  The way to do this is to use a `then:` property - look at Test of Courage for an example.

The reason why it needs to be sequential rather than simultaneous is due to effects like Young Rumormonger.  If you young rumormonger the first honor effect, in your implementation the second will still happen, whereas it shouldn't as the first one didn't resolve successfully.

CodeBaboon:
@jeremylarner  I tried to copy the Test of Courage approach

```
class SoulBeyondReproach extends DrawCard {
    setupCardAbilities(ability) {
        this.action({
            title: 'Honor a character, then honor it again',
            target: {
                cardType: 'character',
                controller: 'self',
                gameAction: ability.actions.honor()
            },
            then: {
                gameAction: ability.actions.honor(context => ({ target: context.target }))
            }
        });
    }
}
```

 but the `then` portion of the ability is not being applied.

```
describe('When successfully played on a character in dishonored state', function() {
    beforeEach(function() {
        this.player1.clickCard(this.soulBeyondReproach);
        this.player1.clickCard(this.hotaru);
        this.player2.clickPrompt('Pass');
    });
    it('should result in the target being in ordinary state and then honored state', function() {
        expect(this.hotaru.isDishonored).toBe(false);
        expect(this.hotaru.isHonored).toBe(false);
        this.player2.clickPrompt('Pass');
        this.player1.clickCard(this.hotaru);
        expect(this.hotaru.isHonored).toBe(true);
    });
});

----------------------------
Failures:
1) Soul Beyond Reproach integration Soul Beyond Reproach's ability When successfully played on a character in dishonored state should result in the target being in ordinary state and then honored state
  Message:
    Expected false to be true.
  Stack:
    Error: Expected false to be true.
```

In trying to figure out what I was doing differently I tried out Test of Courage in-game and it doesn't seem to actually work for that card, either.

image

It might be hard to see, but in that scenario the Scorp player plays Ageless Crone into their Home area during a conflict. Opponent passes. Scorp then uses Test of Courage to bring the Crone into the battle, which should result in an honored Crone, but only the movement happens, not the honoring.

Perhaps this is something wonky on my local copy, so I'll keep attempting to dig into it, but this may be  indicative of a larger issue.

jeremylarner:
Yeah, I'm really sorry - I should have checked that Test of Courage worked before sending you in that direction. Totally my fault.

I've now fixed the bug in Test of Courage (#2072) and the same fix should work for you: you want ```
class SoulBeyondReproach extends DrawCard {
    setupCardAbilities(ability) {
        this.action({
            title: 'Honor a character, then honor it again',
            target: {
                cardType: 'character',
                controller: 'self',
                gameAction: ability.actions.honor()
            },
            then: context => ({
                gameAction: ability.actions.honor({ target: context.target })
            })
        });
    }
}```

CodeBaboon:
Thanks @jeremylarner ! Update coming right up.

PR #2080, opened 2018-06-26
BotK: Heartless Intimidator

### LINKED ISSUE #2024: Heartless Intimidator
image

CDMayberry:
Closes #2024.

Should it display what card was discarded in the effect message?


CDMayberry:
Travis seems to be complaining about me using 'onModifyHonor' in the 'when' property, however this ran fine and the tests worked. Is this a false positive or am I missing something?

PR #2091, opened 2018-06-27
Create ApplyUnopposedAction so it can be interrupted

CodeBaboon:
This is approach to allow for the replacement effect that Ikebana Artisan introduces. From the [developer ruling](http://www.cardgamedb.com/forums/index.php?/topic/39825-ruling-ikebana-artisan/):

> It creates a replacement effect that overrides the normal rule of “unopposed -> lose 1 honor” with a new rule of “unopposed -> lose 1 fate”

By extending `LoseHonorAction` it will continue to default to a 1 honor loss, but by raising the `onApplyUnopposed` event it can now be interrupted in a similar manner to how Display of Power [replaces the default ResolveRingEffect](https://github.com/gryffon/ringteki/blob/a099bd708cc9999ccd8c7e848bacaf0d99c2edf6/server/game/cards/01-Core/DisplayOfPower.js#L13).

### DIFF
--- a/server/game/GameActions/ApplyUnopposedAction.js
+++ b/server/game/GameActions/ApplyUnopposedAction.js
@@ -1,0 +1,20 @@
+const LoseHonorAction = require('./LoseHonorAction');
+
+class ApplyUnopposedAction extends LoseHonorAction {
+    setup() {
+        super.setup();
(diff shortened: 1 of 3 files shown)


jeremylarner:
Closing this as per comments on #2092

Issue #1993, opened 2018-06-20
Ikebana Artisan


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/41689383-3b5e4d90-74e8-11e8-9d3d-c6ef24f0ea89.png)

CodeBaboon:
I'll grab this one.

jeremylarner:
Something to watch with this one, it needs to be able to trigger when the triggering player doesn't have any fate.  That might require some hacks to get around, because I don't think the game will like it.

PR #2092, opened 2018-06-27
TEaF: Ikebana artisan

CodeBaboon:
Implements #1993 
Builds on #2091


jeremylarner:
I realise you're copying another card (Display of Power), but that card uses an ugly hack to deal with the awkward timing issues of that card (it's a reaction to winning a conflict, but creates a delayed cancel).

For this card, I don't think you need quite such a brute force solution.  I'd also prefer to avoid creating new Game Actions which do basically the same thing as old ones. My proposed solution would be as follows:

Modify the current LoseHonorAction to have an dueToUnopposed parameter which defaults to false.
Code Artisan as a would interrupt which triggers on onModifyHonor and looks for that parameter.
Use a handler which cancels the interrupted event, and adds a loseFate event to the window, similar to how Reprieve or Iron Mine work.

CodeBaboon:
@jeremylarner Changes made as per request, ready for re-review.

jeremylarner:
Would you mind merging develop into this branch to make sure it passes tests?

CodeBaboon:
I rebased on master, tests all seem to check out locally:

image

server/game/cards/04.4-TEaF/IkebanaArtisan.js
```diff
@@ -1,10 +1,22 @@
...
+            title: 'Lose fate instead of honor',
+            when: {
+                onModifyHonor: (event, context) => event.dueToUnopposed && event.player === context.player
+            },
+            effect: 'lose 1 fate rather than 1 honor for not defending the conflict',
+            effectArgs: context => context.event.card,
+            handler: context => {
+                context.event.window.addEvent(ability.actions.gainFate({player: context.player, amount: -1}).getEvent(context.player, context));
+                context.cancel();
+            }
```
jeremylarner: This is missing (Unlimited)

CodeBaboon:
Ability is now `unlimited`.

PR #2131, opened 2018-06-28
TL: Press of Battle

### LINKED ISSUE #2105: Press of Battle
image
Water role only. Action: During a MIL conflict,if you control more participating characters than your opponent, choose a non-unique participating character - bow that character.

Kwaice:
Closes #2105


server/game/cards/04.2-TL/PressOfBattle.js
```diff
@@ -1,10 +1,21 @@
 const DrawCard = require('../../drawcard.js');
 
 class PressOfBattle extends DrawCard {
-    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Bow a non-unique character',
+            condition: context => context.game.currentConflict.hasMoreParticipants(context.player) &&
```
jeremylarner: It seems a bit strange to use `context.game` and `this.game` - I would rather you use one or the other (I think I have a slight preference for `this.game` as it's used more in other places).

Also, you need to check `this.game.isDuringConflict('military')` first, as otherwise the game will throw an error if the player clicks the action outside a conflict.

PR #2132, opened 2018-06-28
TFW: Infiltrator

jeremylarner:


jeremylarner:
Needs #2191 to pass tests

PR #2140, opened 2018-06-29
TL: Asahina Takako

jeremylarner:


jeremylarner:
I failed at checking the branch before I committed, so new card implementation got folded into a minor bug fix...

PR #2148, opened 2018-06-29
Test: Shoju and AFWTD

CDMayberry:
We got a report about a 2/2/1 getting hit by Shoju then AFWTD and not discarding. The test I wrote worked without issue, so it doesn't look like their interaction was the problem.


CDMayberry:
Hmm weird, I'll fix this when I get back

PR #2170, opened 2018-07-04
Fix directory structure and change LeavesPlayEvent to work with Tainted Koku

jeremylarner:


NoahTheDuke:
Looks like you missed an `s` in Teachings of the Elements filename

PR #2186, opened 2018-07-04
Test: Court Games

CDMayberry:
Wrote a test for Court Games that includes honoring, based on reports.


CDMayberry:
@NoahTheDuke I re-PR'd, this was accidentally branched from a old card rather than develop.

PR #2217, opened 2018-07-07
TEaF: Heroic resolve

markjbyrne81:


markjbyrne81:
Coveralls check has failed due to HirumaOutpost.js.  Function `leavesPlay()` is not covered.

This confuses me as I didn't change Hiruma Outpost in this PR.

PR #2280, opened 2018-07-16
Test: Chisei

### LINKED ISSUE #2229: chisei district
Hi,
during a game this evening, I was for some reason able to declare a military conflict at the province where chisei district was on as a holding.  The province was unveiled to be rally to the cause, but it shouldn't have mattered as the province should not have been able to be selected to begin with.  
Cheers.

CDMayberry:
Saw #2229, it looks like #2212 fixed the issue, wrote a short test to verify and close #2229. 

Side Note: Don't know if _expect(...).toThrow()_ is the best way to verify, thought it is a way. If anybody has a better idea let me know.


jeremylarner:
It's not the way I would have done it, but I don't know a whole lot about jasmine.

Thanks for doing this! I appreciate someone else following up on bug reports

CDMayberry:
Yep, it's definitely not ideal, catching anything from initiateConflict is pretty broad which isn't great for a test. I'm planning to do Hito District as well, will try to write a more fine-tuned test when I make a pass at it.

We've got a number of people on cards this time around, I figure I can try to keep an eye out for bugs and missing tests.

Issue #2246, opened 2018-07-12
EU: Awakened Tsukumogami


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/42614459-eed391c6-859d-11e8-938c-bc2c748cec00.png)

Issue #2245, opened 2018-07-12
EU: Mantis Tenjinka


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/42614418-c4aad184-859d-11e8-9868-261d6ceb438f.png)

PR #2287, opened 2018-07-18
EU: Mantis Tenkinja & Awakened Tsukumogami

jeremylarner:
Implements #2245 & #2246 
Game engine changes to make the cards work - tests to come later


jeremylarner:
Tests are now added, but I've realised that it's possible to use fate from multiple rings to pay costs with Awakened Tsukumogami if rings gain new elements (due to e.g. Kaede or Seeker of Knowledge).  I'm also uncomfortable with making fate costs uncancellable (especially if players need to manage multiple prompts for choosing fate from rings).  Multiple Tenkinja make cancelling tricky, because once they've been triggered they can't be untriggered

PR #2323, opened 2018-07-21
AaN: Appealing to the fortunes

### LINKED ISSUE #2173: AaN: Appealing to the Fortunes
image

markjbyrne81:
resolves #2173


markjbyrne81:
I based this card off Abandoning Honor but I noticed something weird during testing.

The persistent effect is using the default duration and therefore is only firing once a conflict has bee declared.  From the wording on the card I would think a duration of 'untilEndOfGame' would be better but one does not exist according to the docs.

I thought about using the duration 'untilEndOfRound' but wasnt sure if that would have an unintended side effect.  

I will take a look in the framework tomorrow to try and get to the bottom of it.

jeremylarner:
Not sure what you mean here - persistent effects don't have a duration....

PR #2387, opened 2018-07-31
EU: Roving Michibiku

### LINKED ISSUE #2248: EU: Roving Michibiku
image

CDMayberry:
Closes #2248


CDMayberry:
So technically it doesn't say 'when attacking' as a condition, but I assume claiming one's own ring doesn't change board state.

PR #2398, opened 2018-08-01
Merge pack 6 to master

jeremylarner:


jeremylarner:
Alciende has made some changes to the way data is formatted, and has used a different id for HMT.  Both of those need fixing before merging this.

PR #2402, opened 2018-08-03
Fix: EtV + Legion of One

CDMayberry:
Embrace the Void was preventing Legion of One from resolving it's ability twice because it was technically 'canceling' the event. Added fix and additional tests.


jeremylarner:
I've stolen the tests from here and fixed this bug in #2495 

Closing this so it doesn't cause issues with the new Event system in the PR above

Issue #2421, opened 2018-08-10
Asako Azunami issue


devilsace:
when water ring is claimed and you trigger her, you can bow anyone, but if all your characters are readied, it won't let you pick one and the only way to fix is to go to manuel mode and change it.

CDMayberry:
Got it. I've submitted a fix for this issue.

PR #2422, opened 2018-08-10
Fix for azunami missing cancel prompt

CDMayberry:
The optional flag wasn't actually adding a button to a card select prompt, hence #2421. Should check for manual mode or optional now.


CDMayberry:
Yep, I was considering doing that anyway. Will throw in the changes in a bit.

PR #2430, opened 2018-08-21
Merge develop to master

gryffon:


jeremylarner:
Someone (maybe @CDMayberry?) is adding spaces between `function` and `()`...
I don't really have any preference about which way we do it, but I think it should probably be the same across the codebase and maybe it should be enforced by linting?

CDMayberry:
Ah, yeah I think that's me. VS Codes auto formatter is adding them. I'll fidget with the settings and make it stop.

PR #2432, opened 2018-08-21
quick fix for nezumi leaving play

CDMayberry:
Found the nezumi bug. It starts in `DrawCard.leavesPlay()` on the line `if(this.isDishonored && this.checkRestrictions('affectedByHonor'))`. checkRestrictions isn't getting passed a context and this is causing the crash further down. 

There is a line to check for context but it's happening too late. I moved it back and it's fixed the issue, but I'm a little hesitant to merge it since it seems like a missing context should be an error if anything. Thoughts?

Here's the stack trace.
stack trace

### DIFF
--- a/server/game/cannotrestriction.js
+++ b/server/game/cannotrestriction.js
@@ -30,10 +30,10 @@ class CannotRestriction {
     checkCondition(context) {
         if(!this.restriction) {
             return true;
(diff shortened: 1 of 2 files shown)


jeremylarner:
the checkRestrictions() in DrawCard.leavesPlay() should be passing this.game.getFrameworkContext() as a context.  However, your change is a good idea too

PR #2449, opened 2018-08-29
Fix to allow any # of attachments (and flexbox fixes for hand as well)

Syrail:
Updated attachment code to set style values on attachments (and parent wrapper) to enforce the margin offsets, rather than relying on a pregenerated set of classes. Additionally modified hand code to use flexbox and flex-shrink (with a minimum width of attachment-offset for the current card size) rather than manually setting card positions.

### DIFF
--- a/client/GameComponents/Card.jsx
+++ b/client/GameComponents/Card.jsx
@@ -472,7 +472,7 @@ class Card extends React.Component {
     render() {
         if(this.props.wrapped) {
             return (
-                <div className={ 'card-wrapper' } style={ Object.assign(this.props.style ? this.props.style : {},this.getWrapperStyle()) }>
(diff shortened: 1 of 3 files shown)


client/GameComponents/Card.jsx
```diff
@@ -436,7 +472,7 @@ class Card extends React.Component {
     render() {
         if(this.props.wrapped) {
             return (
-                <div className={ 'card-wrapper ' + this.getWrapper() } style={ this.props.style }>
+                <div className={ 'card-wrapper' } style={ Object.assign(this.props.style ? this.props.style : {},this.getWrapperStyle()) }>
```
jeremylarner: Doesn't this mutate `this.props.style`? I thought that was a no no.... I guess you could fix it by adding `{}` as the first arg in the `assign` function?

Syrail: Good catch! props are supposed to be immutable. Will fix.

PR #2494, opened 2018-09-27
UotE: Asako Maezawa

CDMayberry:


server/game/cards/05-UotE/AsakoMaezawa.js
```diff
@@ -0,0 +1,28 @@
...
+                // My total glory
+                context.player.cardsInPlay.reduce((myTotal, card) => myTotal + (card.isParticipating() && !card.bowed ? card.getGlory() : 0), 0) >
+                // is greater than Opponents total glory
+                context.player.opponent.cardsInPlay.reduce((oppTotal, card) => oppTotal + (card.isParticipating() && !card.bowed ? card.getGlory() : 0), 0)
+            ),
+            target: {
+                cardType: 'character',
+                cardCondition: card => card.isParticipating(),
+                gameAction: ability.actions.cardLastingEffect(context => ({
+                    effect: ability.effects.modifyBasePoliticalSkill(context.target.getBasePoliticalSkill())
```
jeremylarner: I think this maybe should be a multiplier, not a modifier.... It's worded the same as Vengeful Berzerker

jeremylarner: Actually, scratch that, I think that's a can of worms which we don't want to open, and until they print an ability which has an additive effect on base skill (rather than a set, switch or multiplier), I'd rather not mess with it....

PR #2510, opened 2018-10-08
UotE: Make An Opening

### LINKED ISSUE #2502: UotE: Make an Opening
image

CodeBaboon:
Closes #2502

### DIFF
--- a/server/game/cards/05-UotE/MakeAnOpening.js
+++ b/server/game/cards/05-UotE/MakeAnOpening.js
@@ -1,0 +1,34 @@
+const DrawCard = require('../../drawcard.js');
+
+class MakeAnOpening extends DrawCard {
+    setupCardAbilities(ability) { // eslint-disable-line no-unused-vars
+        this.action({
+            title: 'Give -X/-X to opposing character, where X is the difference between current honor dial bid values',
+            condition: context =>
+                this.game.isDuringConflict() &&
+                this.game.currentConflict.getNumberOfParticipantsFor(context.player) >= 1 &&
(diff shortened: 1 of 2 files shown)


server/game/cards/05-UotE/MakeAnOpening.js
```diff
@@ -0,0 +1,34 @@
...
+                this.game.currentConflict.getNumberOfParticipantsFor(context.player.opponent) >= 1 &&
+                context.player.showBid &&
+                context.player.opponent.showBid &&
+                context.player.showBid !== context.player.opponent.showBid,
+            target: {
+                cardType: 'character',
+                controller: 'opponent',
+                cardCondition: card =>
+                    card.isParticipating() &&
+                    (card.militarySkill > 0 || card.politicalSkill > 0),
```
jeremylarner: It's legal to play this on a 0/0 character, as it's still considered a change in game state, so you shouldn't check for positive skills here

server/game/cards/05-UotE/MakeAnOpening.js
```diff
@@ -0,0 +1,34 @@
...
+                gameAction: ability.actions.cardLastingEffect(context => ({
+                    effect: ability.effects.modifyBothSkills(() => -(this.getHonorDialDifference(context)))
+                }))
```
jeremylarner: This should be a static effect, not a dynamic one I think.  The value of X is locked in when the ability resolves - if dials change later in the conflict, the bonus shouldn't change

PR #2511, opened 2018-10-09
UotE: Stolen Secrets

jeremylarner:
Also contains some changes to PlayableLocation allowing limitation to certain cards in that location which can be played


jeremylarner:
I have no idea why this is changing test coverage so much

PR #2523, opened 2018-10-11
UtoE: From The Shadows

### LINKED ISSUE #2469: UotE: From the Shadows
shadows

CodeBaboon:
Closes #2469

### DIFF
--- a/server/game/Events/EntersPlayEvent.js
+++ b/server/game/Events/EntersPlayEvent.js
@@ -6,8 +6,7 @@ class EntersPlayEvent extends Event {
         this.handler = this.entersPlay;
         this.card = card;
         this.fate = fate;
-        this.card.isHonored = status === 'honored';
-        this.card.isDishonored = status === 'dishonored';
+        this.status = status;
         this.gameAction = gameAction;
         this.originalLocation = card.location;
     }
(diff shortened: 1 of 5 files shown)


test/server/cards/05-UotE/FromTheShadows.spec.js
```diff
@@ -1,29 +1,140 @@
...
+                            expect(this.disguisedProtector.isDishonored).toBe(true);
+                        });
+
+                        describe('if Young Rumormonger is in play', function() {
+                            beforeEach(function() {
+                                this.youngRumormonger = this.player1.placeCardInProvince('young-rumormonger', 'province 1');
+                                this.player1.putIntoPlay(this.youngRumormonger);
+                            });
+
+                            it('should allow Young Rumormonger to use his ability', function() {
```
CodeBaboon: This should say 'not'

server/game/Events/EntersPlayEvent.js
```diff
@@ -1,11 +1,13 @@
...
 
 class EntersPlayEvent extends Event {
-    constructor(params, card, fate, gameAction) {
+    constructor(params, card, fate, status, gameAction) {
         super('onCharacterEntersPlay', params);
         this.handler = this.entersPlay;
         this.card = card;
         this.fate = fate;
+        this.card.isHonored = status === 'honored';
+        this.card.isDishonored = status === 'dishonored';
```
jeremylarner: We can't change the card state in the constructor, because if something cancels this event, we'll end up with an honored card in hand/province.  Changing the state needs to be in the handler.  I also think it might be better to honor/dishonor by calling the honor/dishonor method on the card.

CodeBaboon: Good catch, I suppose I should be at least doing it in the same place the `card.fate` is changed down below. I'll try calling it in via the `dishonor` method on the card but I did that briefly yesterday and it didn't work (might have just been PEBKAC).

Issue #2560, opened 2018-10-19
Peasants Advice Reveals Face Down Provinces in Character Selection Prompt


Fite-Ian:
When playing Peasants Advice, a character must be chosen to be dishonored and a province to be peeked at.  Like with most events, it's possible to cancel playing PA before selecting a character to target, however after a province has been targeted but before a character is selected the name of the province is included in the prompt even if it is face down.  It is possible to initiate and cancel playing PA to see the names of all of the opponent's provinces without actually playing it.

jamesoddy:
Confirmed.  Screenshot:

![image](https://user-images.githubusercontent.com/958931/47207661-b3789200-d383-11e8-96da-cfb3d2652df0.png)

Opportunity to cancel by opponent also does not appear until after the province has been revealed.

PR #2561, opened 2018-10-19
Hide facedown targeted cards

jamesoddy:
Partial fix for #2560 

Not a perfect fix, but hides facedown cards when chosen as a target.  Ideally, the controller should be able to see their own facedown provinces when chosen as a target, but I could not work out a way to get the controller of the target to be passed to the AbilityTargeting.jsx file.

### DIFF
--- a/client/GameComponents/AbilityTargeting.jsx
+++ b/client/GameComponents/AbilityTargeting.jsx
@@ -4,7 +4,7 @@ import _ from 'underscore';
 
 class AbilityTargeting extends React.Component {
     onMouseOver(event, card) {
-        if(card && this.props.onMouseOver) {
+        if(card && !card.facedown && this.props.onMouseOver) {
             this.props.onMouseOver(card);
         }
     }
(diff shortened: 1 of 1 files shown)


jeremylarner:
What does this look like in game?  My major concern (though I haven't spent much time looking at the problem, so I might be completely wrong) is that you're fixing the problem by changing what the client displays to the player, whereas the problem here is that the server is sending information the player shouldn't have to the client.  My gut feeling is that this problem needs to be fixed at the server level (it might need client fixes too)

jeremylarner:
#2563 is my attempt to fix this bug - what do you think @jamesoddy ?

jamesoddy:
@jeremylarner I did have a bit of a look last night.  Although the facedown image is shown correctly in the targeting prompt - I couldn’t get the card zoom to display correctly (for either player). As far as the card zoom was concerned the card was “undefined”. I ran out of time before I could investigate further. That being said certainly seemed a more robust solution than my attempt!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jamesoddy:
@jeremylarner sorry just saw this.  This is what I am getting when I run it on my own system:

image

The top right image is missing because it is looking for /img/cards/undefined.jpg

image

This is both for the player of peasants advice and also the opponent.  Ideally, the opponent should be able to see the facedown card from the ability targeting prompt and the player of peasant's advice should not be able to (in this case).  So the card zoom is not honoring the "only look at a facedown card if you are allowed to" and it is seeing the card as undefined.

I'll try to have a look more later to see if I can help any further.

jeremylarner:
Ok, I'm pretty confident I've fixed the bug in the other PR - closing this one

PR #2563, opened 2018-10-20
Fix for facedown cards in targeting controls

### LINKED ISSUE #2560: Peasants Advice Reveals Face Down Provinces in Character Selection Prompt

jeremylarner:
Fixes #2560


jamesoddy:
Looks to work perfectly now!

PR #2568, opened 2018-10-24
fetchdata.js - handle cards with mutliple cards in pack

jamesoddy:
Images were not being fetched for any cards that had mutliple cards in it's pack (such as way-of-the-crane that has a promo card from Gen Con 2017).

### DIFF
--- a/server/scripts/fetchdata.js
+++ b/server/scripts/fetchdata.js
@@ -41,23 +41,13 @@ let fetchCards = apiRequest('cards')
 
         var i = 0;
 
-        cards.forEach(function(card) {
-            var imagePath = path.join(imageDir, card.id + '.jpg');
-            let imagesrc;
-
-            if(card.pack_cards.length === 1 && card.pack_cards[0].image_url) {
-                imagesrc = card.pack_cards[0].image_url;
-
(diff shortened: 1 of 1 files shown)


server/scripts/fetchdata.js
```diff
@@ -41,23 +41,13 @@ let fetchCards = apiRequest('cards')
...
-            if(card.pack_cards.length === 2 && card.pack_cards[1].image_url) {
-                imagesrc = card.pack_cards[1].image_url;
-
-                if(imagesrc && !fs.existsSync(imagePath)) {
-                    fetchImage(imagesrc, card.id, imagePath, i++ * 200);
+        cards.forEach(function (card) {
+            if(card.pack_cards.length > 0) {
+                var imagePath = path.join(imageDir, card.id + '.jpg');
+                let firstCardWithImageUrl = card.pack_cards.find(card => card.image_url);
+                let imageSrc = firstCardWithImageUrl.image_url;
```
jeremylarner: won't this throw an error if the find returns undefined?

jamesoddy: good catch - I have updated with a new commit

Issue #2415, opened 2018-08-08
Graphical Bug - All rings turn political while opponent is selecting a ring


YogoGohei:
![image](https://user-images.githubusercontent.com/32277970/43846787-2e20035e-9afd-11e8-9c96-fca8ae4022a6.png)

This can be confusing when cards like Shiotome Encampment are in play and you are trying to plan ahead.

In the attached image, "Unseen" had claimed earth as a military ring.

YogoGohei:
This isn't just during ring selection, it's during the entire conflict.  The selected ring is bright and displayed correctly, but all non-selected rings fade to political.  This is a problem when trying to assess game state (again, Shiotome Encampment being the big offender).

YogoGohei:
This looks great and has worked well in the games since it was implemented.  Thank you!

PR #2570, opened 2018-10-24
Display conflict type icon for claimed/contested rings

jamesoddy:
For issue #2415


jeremylarner:
When you're making client changes like this, it would be useful to have a screenshot showing what you've done

jamesoddy:
Of course, this adds little conflict type icons to contested and claimed rings:

image

This is because during a conflict or ring selection the rings are faded so you cannot tell if players have claimed military or political rings (for purposes of cards such as shiotome encampment).

More details in issue #2415

Issue #2416, opened 2018-08-08
Graphical Bug - Ring images broken when selecting effect


YogoGohei:
![image](https://user-images.githubusercontent.com/32277970/43846857-585ba3f8-9afd-11e8-8cd6-99437c3a8d7b.png)

jamesoddy:
Fixed in PR #2575 

Now shows as:

![image](https://user-images.githubusercontent.com/958931/47907782-ca3ede80-de83-11e8-8041-aa14d58965e7.png)

PR #2575, opened 2018-11-02
Added missing conflictType to short summary for rings

jamesoddy:
Fix for issue #2416.

Let's just pretend it didn't take me hours to find the problem ;)


jamesoddy:
For reference the ring effect prompt should now look like this:

image

Instead of this:

image

PR #2577, opened 2018-11-02
Update dependencies to work with node v10.

fatihi:


fatihi:
Will move to a branch and open again

PR #2579, opened 2018-11-03
Young Rumormonger bugfix

### LINKED ISSUE #2576: Master Alchemist didn't trigger Young Rumermonger
image

Master Alchemist was used to change Togashi Mitsu from dishonored to normal.
Young Rumormonger was not given a chance to trigger.
Unless I am missing some ruling somewhere, I think I can move that to someone else.  He did have another normal character to move it to.

jeremylarner:
Fixes #2576


jeremylarner:
I'm not 100% happy with this fix.  It works, but I think the bug points to an issue with what is currently fireRingAction.  It probably should be a chooseAction rather than it's own thing.

PR #2588, opened 2018-11-05
Tests - All and Nothing

jamesoddy:
100% coverage for All or Nothing (Elemental Cycle)

### DIFF
--- a/test/server/cards/04.5-AaN/AllAndNothing.spec.js
+++ b/test/server/cards/04.5-AaN/AllAndNothing.spec.js
@@ -6,7 +6,6 @@ describe('All and Nothing', function() {
                     phase: 'conflict',
                     player1: {
                         inPlay: ['isawa-kaede', 'seeker-of-knowledge'],
-                        // hand: ['all-and-nothing', 'all-and-nothing']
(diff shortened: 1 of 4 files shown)


test/server/cards/04.5-AaN/AllAndNothing.spec.js
```diff
@@ -6,8 +6,8 @@ describe('All and Nothing', function() {
                     phase: 'conflict',
                     player1: {
                         inPlay: ['isawa-kaede', 'seeker-of-knowledge'],
-                        hand: ['all-and-nothing', 'all-and-nothing']
-                        // hand: ['all-and-nothing', 'all-and-nothing', 'taryu-jiai']
+                        // hand: ['all-and-nothing', 'all-and-nothing']
```
NoahTheDuke: Why not just remove the whole line?

jamesoddy: Yeah I noticed this after I setup the PR and didn't think it was worth adding another commit.  Just something I missed.  Completely agree that the commented out line should be removed.

PR #2599, opened 2018-11-08
Duels - Tests

jamesoddy:
Includes PR #2598 - Bug fix for duels where one participant dies.

Multiple tests for all existing duels particularly when the duel target is discarded during the duel.


server/game/cards/01-Core/KakitaKaezin.js
```diff
@@ -24,7 +24,7 @@ class KakitaKaezin extends DrawCard {
         if(winner === context.source) {
             this.game.addMessage('{0} wins the duel, and sends all characters except {0} and {1} home', winner, loser);
             this.game.applyGameAction(context, { sendHome: this.game.currentConflict.getParticipants(card => ![winner, loser].includes(card)) });
-        } else if(loser === context.source) {
+        } else {
```
jeremylarner: Won't this throw an error if Kaezin dies during the duel?

PR #2622, opened 2018-11-09
Tests for CotE

jamesoddy:
I will use this branch to start writing tests for CotE.

The tests will be disabled at first using `xdescribe` instead of `describe` otherwise the build will fail, but this will allow people to bring the tests in to test against implementation.

Please bear in mind that I am writing these tests blind so they are likely to not be perfect and will likely need adjustment once the implementation is complete, but they should act as a starter at least.


NoahTheDuke:
This is some wild test driven development, haha. I love it

jamesoddy:
No longer relevant.  PR #2627 & PR #2625 contain tests and implementation.

PR #2627, opened 2018-11-13
CotE - Courtly Challenger

jamesoddy:
Includes commit bb2ae68 (Bug fix for duels where one participant dies).

Tests run ok if courtly-challenger exists ;)

### DIFF
--- a/server/game/cards/06-CotE/CourtlyChallenger.js
+++ b/server/game/cards/06-CotE/CourtlyChallenger.js
@@ -7,7 +7,7 @@ class CourtlyChallenger extends DrawCard {
             effect: [
                 ability.effects.delayedEffect({
                     when: {
-                        afterDuel: (event) => event.winner && event.winner === this
+                        afterDuel: (event, context) => event.winner && event.winner === context.source
                     },
                     multipleTrigger: true,
                     message: '{0} is honored due to winning a duel',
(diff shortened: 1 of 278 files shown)


server/game/cards/06-CotE/CourtlyChallenger.js
```diff
@@ -0,0 +1,51 @@
...
+
+class CourtlyChallenger extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            effect: [
+                ability.effects.delayedEffect({
+                    when: {
+                        afterDuel: (event) => event.winner && event.winner === this
+                    },
+                    message: '{0} is honored due to winning a duel',
```
jeremylarner: This should have `multipleTrigger: true`

server/game/cards/06-CotE/CourtlyChallenger.js
```diff
@@ -0,0 +1,51 @@
...
+                    gameAction: ability.actions.honor(context => ({ target: context.source }))
+                })
+            ]
+        });
+        this.persistentEffect({
```
jeremylarner: Is there a specific reason for separating the two persistent effects? I'd prefer them to be together, just in case there is ever an ability which copies a specific constant ability

server/game/cards/06-CotE/CourtlyChallenger.js
```diff
@@ -0,0 +1,51 @@
...
+                    when: {
+                        afterDuel: (event) => event.loser && event.loser === this
+                    },
+                    message: '{0} is dishonored due to losing a duel',
+                    gameAction: ability.actions.dishonor(context => ({ target: context.source }))
+                })
+            ]
+        });
+        this.action({
+            title: 'Challenge a participating character to a Political duel: the winner draws 2 cards',
```
jeremylarner: This title is probably too long, it needs to fit nicely on a prompt button.  It doesn't need to give so many details, just to give the player enough to distinguish it from other abilities (e.g. Duelist Training)

Issue #2603, opened 2018-11-09
CotE : Honest Challenger


jamesoddy:
![Honest Challenger](https://cdn.discordapp.com/attachments/316605644579012608/499597318740246529/l5c16-a1-honest-challenger.png)

PR #2629, opened 2018-11-14
CotE - Honest Challenger

jamesoddy:
For issue #2603


jamesoddy:

PR #2633, opened 2018-11-15
CotE - Hand to Hand

### LINKED ISSUE #2609: CotE : Hand to Hand
Hand to Hand

jamesoddy:
Closes #2609 

Not perfect.  Currently it still prompts even when there are no more valid targets, I'd rather it would auto finish at that point.  I haven't the time to look at it at the moment, but wanted to at least get this part out there.  When I get the chance I may have another look.

### DIFF
--- a/server/game/GameActions/ResolveAbilityAction.js
+++ b/server/game/GameActions/ResolveAbilityAction.js
@@ -38,7 +38,14 @@ class ResolveAbilityAction extends CardAction {
     }
 
     canAffect(card, context) {
-        return this.ability && super.canAffect(card, context);
+        if(!super.canAffect(card, context) || !this.ability || !this.secondResolution && context.player.isAbilityAtMax(this.ability.maxIdentifier)) {
+            return false;
+        }
+        let newContext = this.ability.createContext(context.player);
+        if(this.ability.targets.length === 0) {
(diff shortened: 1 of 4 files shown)


jeremylarner:
There are a couple of issues with the way you are doing this.  Firstly, I think it would be better to just have a single then, which resolves the ability again for the other player (I'm not sure whether we have the infrastructure to do that, but I think it's the best approach).  Secondly, you're using `promptForSelect` somewhere where it shouldn't really be used.  The intention for that property is that it's used when resolution requires decisions to be made.  Here, we go through another targeting step, so just giving a `target` property to the then is the best way to do it.

jamesoddy:
I went through several iterations to get to this and I was expecting some negative feedback - that's partly why I wanted to get it out there to give you a chance to take a look.

I had it working with a single `then` by adding a `player` field to the `ResolveAbilityAction` but the issue was that I could not find an `optional` field so there is no way for someone to end the cycle that's why I ended up with using `promptForSelect`.

Please by all means pick it apart and come up with a cleaner solution, but this is at least a passing solution for the time being.

jamesoddy:

jamesoddy:
Merged in PR #2660 and adjusted implementation and tests to take advantage of the ReolveAbilityAction now checking it can resolve.

PR #2638, opened 2018-11-18
Fix bug with province refills

jeremylarner:


jamesoddy:
So is this to basically now "refill" after every game action?

Issue #2610, opened 2018-11-09
CotE : Karmic Balance


jamesoddy:
![Karmic Balance](https://cdn.discordapp.com/attachments/316605644579012608/502186783942115328/Definitely_not_a_Spoiler.png)

PR #2639, opened 2018-11-18
CotE - karmic balance

jamesoddy:
Definitely needs review this one.

I believe this is the first event that removes itself from the game.  This caused me some issues as the event card is normally moved to the discard pile *after* the ability is resolved.

I solved this by adding a new property to `CardAbility` that is then checked in `abilityresolver` to "removeFromGame" instead of moving to the discard pile.

I also think the `returnToDeck` action could be used here, but I had trouble getting that to work with the hand/conflict discard so any help here would be appreciated.

#2610


jeremylarner:
I haven't had chance to look at this in a lot of detail, but two initial thoughts:
1) I'd prefer to avoid `handler` as much as possible.  For effects which we don't have game actions for (and for which it's difficult to make game actions for) it's an option, but I think you can do this card with current game actions (did you try moveCard?)
2) I can see why you got to where you did with removeFromGame, but another approach would be to make the move-to-discard function in abilityresolver.js check that the card is still in 'being resolved' before moving it? I'm not sure if that's a better approach or not, but it feels more flexible to me.

jamesoddy:
1.  Yes, I suspect there is a way to avoid using handler, but I struggled and caved in to just using handler.  From what I can see moveCard and returnToDeck are both only designed to work with a single target so would need modifying to deal with this.  I will have another go at some point.
2.  That seems a much simpler and, as you say, flexible solution.  I've tried it out and works well.

jeremylarner:
I think moveCard will work fine with multiple cards.  The only issue I see is that it shuffles after every card is moved.  I think this could probably be remedied reasonably easily though...

jamesoddy:
Great.  I must be misunderstanding how the abilites are called then so I will take another look.

Shuffling after every card should just make it really well shuffled :wink:

Just realised I am also only doing this for the source.player as well so need to adjust tests and implementation for both players.  :smile:

jeremylarner:
The shuffling thing is only a problem in that it will spam the chat log with 'X is shuffling their conflict deck'

jamesoddy:
Whoops.  I've committed more that I should have.  I will close this PR and start a new one.  Gives me a chance to squash some of these commits as well.

jamesoddy:
Closes #2610

server/game/GameActions/ReturnToDeckAction.js
```diff
@@ -34,7 +34,7 @@ class ReturnToDeckAction extends CardGameAction {
                 event.context.refillProvince(card.controller, card.location, context);
             }
             card.owner.moveCard(card, destination, { bottom: this.bottom });
-            if(this.shuffle) {
+            if(this.shuffle && (card === this.target[this.target.length - 1])) {
```
jeremylarner: This won't work if getEvent is called outside of the context of an ability (which happens in some places I think)

jeremylarner: yes, that's fine

jamesoddy:

PR #2642, opened 2018-11-21
Tests - Bayushi's Whisperers

jamesoddy:
In the process of implementing Akodo Toturi, noticed that Bayushi's Whisperer's was not honoring the cannot restriction on playing characters.

This fixes that and adds a test for it (as well as events and attachments).


jamesoddy:
Includes PR #2652 that fixes canPlay so this now just includes some extra tests for Bayushi's Whisperers.

PR #2643, opened 2018-11-21
CotE - Akodo Toturi

### LINKED ISSUE #2607: CotE : Akodo Toturi
akodo-toturi

jamesoddy:
Includes bugfix PR #2642

Closes #2607

Added new 'fromHand' cannotRestriction, please reivew if this is the best way of dealing with this effect.


jamesoddy:
Ok, just realised I didn't check if this works with Hidden Moon Dojo, etc.

Looks like it doesn't so I need to write some more tests and changes.

jamesoddy:
Ok, just to clarify what I have done.

In order to restrict playing from hand I have added `playType` to `AbilityContext` this is then set in the various appropriate actions to be 'playFromHand'.  Then Toturi simply adds a new type of restriction of `play` from `fromHand` that then checks the `playType` in the context.  If it is 'playFromHand' then the trigger is blocked.

Hope that's clear.

Toturi has fairly extensive tests, but I did not cover every eventuality.  The theory is if Mitsu is blocked then Kyuden Isawa should be blocked as they use the same mechanism.

PR #2645, opened 2018-11-23
Bugfix - Stolen Secrets

jamesoddy:
Do not merge!  This doesn't fix the problem.

Added multiple new tests and tried to set as facedown when removed from play, but that seems to break a lot of things!


jamesoddy:
Closes #2654

jamesoddy:
I have tidied up and set to your implementation of using the hideWhenFaceup effect as requested.

PR #2655, opened 2018-11-28
CotE - Heavy Ballista

### LINKED ISSUE #2613: CotE : Heavy Ballista
Heavy Ballista

jamesoddy:
Closes #2613


jamesoddy:
It's a bit pendantic, but the prompted player should be the controller of the targeted unit (not necessarily the opponent).

In order to achieve this, I have added the ability to specify the player by function in `AbilityTargetSelect`.  I think I have done this correctly (all tests pass), but could do with review.

jeremylarner:
I'm struggling to think of a scenario where you could be both the player initiating this ability and control an attacking character.  I'm not opposed to adding support for a function in `player:`, but I think it should match the format in AbilityTargetCard.  I think there the function returns a constant, whereas in yours it returns a Player object.  This points towards the fact that there should probably be a class under each kind of AbilityTarget which they inherit from....

jamesoddy:
Yeah I don't think the scenario can happen yet, but there may be some weird way in the future.  I have re-implemented in the same way as `AbilityTargetCard` as requested.

As you say a common ancestor for the AbilityTarget classes would probably be a good idea at some time.

PR #2656, opened 2018-11-29
CotE - Sage of Gisei Toshi

### LINKED ISSUE #2612: CotE : Sage of Gisei Toshi
Sage of Gisei Toshi

jamesoddy:
Closes #2612

### DIFF
--- a/server/game/cards/06-CotE/SageOfGiseiToshi.js
+++ b/server/game/cards/06-CotE/SageOfGiseiToshi.js
@@ -1,0 +1,24 @@
+const DrawCard = require('../../drawcard.js');
+const { Players, CardTypes } = require('../../Constants');
+
+class SageOfGiseiToshi extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Move home, then move character home',
+            condition: context => this.game.isDuringConflict() && context.player.opponent && context.player.opponent.honor < context.player.honor,
+            gameAction: ability.actions.sendHome(),
+            target: {
+                cardType: CardTypes.Character,
+                controller: Players.Opponent,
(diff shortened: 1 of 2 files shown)


server/game/cards/06-CotE/SageOfGiseiToshi.js
```diff
@@ -0,0 +1,24 @@
+const DrawCard = require('../../drawcard.js');
+const { Players, CardTypes } = require('../../Constants');
+
+class SageOfGiseiToshi extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Move home, then move character home',
+            condition: context => this.game.isDuringConflict() && context.player.opponent && context.player.opponent.honor < context.player.honor,
```
jeremylarner: This ability targets before resolving, so the target: should be here, not in the then

jamesoddy: Gotcha, I tried it like that at first and it gave me a bit of trouble, but think I've sussed it now.  I've pushed a new version.

server/game/cards/06-CotE/SageOfGiseiToshi.js
```diff
@@ -0,0 +1,24 @@
...
+            gameAction: ability.actions.sendHome(),
+            target: {
+                cardType: CardTypes.Character,
+                controller: Players.Opponent,
+                cardCondition: (card) => card.isParticipating()
```
jeremylarner: I think we need to check `card.allowGameAction('moveHome', context)` here as well. Normally the game will do that automatically, but I don't think it will where the target is only affected by the post-then part of the effect

jamesoddy: Done.

PR #2675, opened 2018-12-10
Embrace the Void / Karmic Twist / Feast or Famine Fix

### LINKED ISSUE #2640: feast or famine and embrace the void issue
i broke my opponent's feast or famine. i had embrace the void on tadaka. they used feast, i then used embrace.  tadaka did not lose the fate. my opponent also had guest of honor and menacing iron warrior  out and i had asako azunami and isawa uona out. the conflict was a water ring.  asako azunami was bowed from defending and winning the fire ring, the previous conflict.  

i'm giving the names of the other cards just in case it had something to do with affecting the coding.

jamesoddy:
Fixes #2640 

Potential re-working of Karmic Twist and fix for Feast or Famine/Embrace the Void interaction.

Allows PlaceFateAction to now target a player instead of just a card so that Embrace the Void can replace the target with a player.

### DIFF
--- a/server/game/Events/MoveFateEvent.js
+++ b/server/game/Events/MoveFateEvent.js
@@ -1,4 +1,5 @@
 const Event = require('./Event.js');
+const { CardTypes } = require('../Constants');
 
 class MoveFateEvent extends Event {
     constructor(params, fate, origin, recipient, gameAction) {
@@ -8,6 +9,23 @@ class MoveFateEvent extends Event {
         this.recipient = recipient;
         this.fate = fate;
         this.gameAction = gameAction;
+    }
+
+    checkCondition() {
+        if(this.cancelled || this.resolved) {
+            return;
+        }
+
+        if(this.origin) {
+            if(this.origin.fate === 0 || this.origin === CardTypes.Character && !this.origin.allowGameAction('removeFate', this.context)) {
+                this.cancel();
+                return;
+            }
+        }
+
+        if(this.recipient && this.recipient === CardTypes.Character && !this.recipient.allowGameAction('placeFate', this.context)) {
+            this.cancel();
+        }
     }
 
     moveFate() {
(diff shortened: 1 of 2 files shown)


server/game/GameActions/PlaceFateAction.js
```diff
@@ -6,19 +6,27 @@ class PlaceFateAction extends CardGameAction {
     setDefaultProperties() {
         this.amount = 1;
         this.origin = null;
+        this.recipient = null;
```
jeremylarner: Using `this.recipient` here doesn't make any sense to me, because the target of the ability should be the recipient.  You don't actually use this parameter when generating the event which actually changes the game state

server/game/GameActions/PlaceFateAction.js
```diff
@@ -6,19 +6,27 @@ class PlaceFateAction extends CardGameAction {
...
     }
 
     setup() {
         this.name = 'placeFate';
-        this.targetType = [CardTypes.Character];
+        this.targetType = [CardTypes.Character, 'player'];
```
jeremylarner: I'm not particularly happy about this change because it means the game will let you place fate on a player, which shouldn't really be possible.  I think this change alone should fix all the issues though

server/game/GameActions/PlaceFateAction.js
```diff
@@ -6,19 +6,27 @@ class PlaceFateAction extends CardGameAction {
...
+
+    checkRecipient(card, context) {
+        if(!this.recipient || ['player', 'ring'].includes(this.recipient.type)) {
+            return true;
+        }
+        if(this.amount === 0 || (card && card.location !== Locations.PlayArea)) {
             return false;
         }
-        return super.canAffect(card, context) && this.checkOrigin(context);
+        return this.recipient.allowGameAction('placeFate', context);
```
jeremylarner: I'm not really sure what the changes to `canAffect` and adding `checkRecipient` are intended to achieve...

PR #2678, opened 2018-12-11
Update the restricted list to 1.7

jeremylarner:


NoahTheDuke:
This doesn't matter a whole lot, but is there any reason to not sort the list?

jeremylarner:
It's in the same order as in the RRG.  I thought that would make it easier to check for changes, or to make changes

NoahTheDuke:
That makes sense, thanks!

PR #2683, opened 2018-12-11
Bugfix for Feast or Famine/Embrace the Void interaction

jeremylarner:
Also cleans up Karmic Twist


jeremylarner:
Incorporated into #2675

Issue #2619, opened 2018-11-09
CotE: The Fires of Justice


Kwaice:
The Fires of Justice - Lion Clan Event

2 Fate, 2 Influence

Void role only. Reaction: After you win a ? conflict, your opponent chooses a participating character he or she controls. Select one – remove all fate from that character, or move any amount of fate from your opponent’s fate pool onto that character.

PR #2690, opened 2018-12-13
CotE - The Fires of Justice

jamesoddy:
Relates to #2619.

I may be going about this the completely wrong way and this currently does not work.

The implementation is quite simple until you get to the "move any amount of fate from your opponent’s fate pool onto that character".  As it is any amount, we need to prompt to allow the player to select any amount up to the opponent's fate amount.  As this could be quite high I figured I could re-tool the code from `costs.js` when choosing the additional fate as there may not be enough room to display all amounts.

I decided to put the prompts as part of MoveFateEvent so it would be re-usable in future.  It appears that the prompts are working ok, but the fate is never transfered (as `context.chooseFate` remains as zero).  I'm sure this is due to my ignorance of how the context is being passed through the handlers.

Any help would be much appreciated.


server/game/cards/06-CotE/TheFiresOfJustice.js
```diff
@@ -0,0 +1,46 @@
...
+                    cardCondition: card => card.isParticipating()
+                },
+                select: {
+                    mode: TargetModes.Select,
+                    dependsOn: 'character',
+                    choices: {
+                        'Remove all fate': AbilityDsl.actions.removeFate(context => ({ target: context.targets.character, amount: context.targets.character.fate })),
+                        'Move fate to character': AbilityDsl.actions.menuPrompt(context => ({
+                            activePromptTitle: 'Select fate amount:',
+                            choices: Array.from(Array(context.player.opponent.fate), (x, i) => (i + 1).toString()),
```
jeremylarner: You change this to an array of strings - to you need to turn it back to an int when you pass it as a property in `choiceHandler`?

PR #2691, opened 2018-12-13
WotW - Iuchi Daiyu

### LINKED ISSUE #2666: WotW: Iuchi Daiyu
image

jamesoddy:
Closes #2666

Includes a helper function on `player.js` to get the number of faceup non-stronghold provinces a player controls.

Needs the cards to be on fiveringsdb-data before tests will run successfully.  I'll get on that.


server/game/cards/07-WotW/IuchiDaiyu.js
```diff
@@ -0,0 +1,21 @@
+const DrawCard = require('../../drawcard.js');
+
+class IuchiDaiyu extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: '+1 military for each faceup province',
+            condition: () => this.game.isDuringConflict(),
```
jeremylarner: This should probably check for an opponent too.

Alternatively, you could make the helper function return the number of faceup provinces their opponent controls (0 if no opponent)

jamesoddy: I have added a new function `getNumberOfOpponentsFaceupProvinces(predicate)` that will return 0 if opponent is not defined.

PR #2702, opened 2018-12-17
WotW - Shinjo Trailblazer

### LINKED ISSUE #2651: CotE: Shinjo Trailblazer
Shinjo Trailblazer

jamesoddy:
Closes #2651

### DIFF
--- a/server/game/cards/07-WotW/ShinjoTrailblazer.js
+++ b/server/game/cards/07-WotW/ShinjoTrailblazer.js
@@ -1,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class ShinjoTrailblazer extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            title: 'Gain +2/+2',
+            when: {
+                onCardRevealed: (event, context) => event.card.isProvince && event.card.controller === context.source.controller.opponent && this.game.isDuringConflict()
+            },
+            gameAction: ability.actions.cardLastingEffect({ effect: ability.effects.modifyBothSkills(2) }),
+            effect: 'to give {0} +2{1}, +2{2}',
+            effectArgs: () => ['military', 'political']
+        });
+    }
+}
+
+ShinjoTrailblazer.id = 'shinjo-trailblazer';
+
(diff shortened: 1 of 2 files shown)


server/game/cards/07-WotW/ShinjoTrailblazer.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+
+class ShinjoTrailblazer extends DrawCard {
+    setupCardAbilities(ability) {
+        this.reaction({
+            title: 'Gain +2/+2',
+            when: {
+                onCardRevealed: (event, context) => event.card.isProvince && event.card.controller === context.source.controller.opponent && this.game.isDuringConflict()
```
jamesoddy: Please confirm if this is the best way to check if it's an opponent's province.  Seemed a bit wordy to me.

server/game/cards/07-WotW/ShinjoTrailblazer.js
```diff
@@ -0,0 +1,19 @@
...
+            },
+            gameAction: ability.actions.cardLastingEffect({ effect: ability.effects.modifyBothSkills(2) }),
+            effect: 'to give {0} +2{1}, +2{2}',
+            effectArgs: () => ['military', 'political']
```
jeremylarner: You don't need to do this any more - the new chat log looks for military and political within string text

server/game/cards/07-WotW/ShinjoTrailblazer.js
jeremylarner: I think that's the best way to do it....

server/game/cards/07-WotW/ShinjoTrailblazer.js
jamesoddy: So just to clarify we would just have

```
effect: 'to give {0} +2military, +2political',
```
with no spaces?

jeremylarner: Huh, you're right I hadn't thought of that.  It will  only replace them as separate words (i.e. with spaces on both sides...)

PR #2729, opened 2019-01-02
CotE - Honorable Challenger

### LINKED ISSUE #2706: CotE - Honorable Challenger
image

jamesoddy:
Closes #2706 

As far as I can see we've not had a lasting effect from a duel before so not 100% sure how it needed to be implemented.  I had to add some extra properties to the interface, but this passes my tests.

Could do with review though to check I am going about this correctly (and adding to the interface is correct).

### DIFF
--- a/server/game/cards/06-CotE/HonorableChallenger.js
+++ b/server/game/cards/06-CotE/HonorableChallenger.js
@@ -1,0 +1,23 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl');
+
+class HonorableChallenger extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Initiate a military duel',
+            initiateDuel: context => ({
+                type: 'military',
+                resolutionHandler: (winner) => this.resolutionHandler(context, winner)
+            })
+        });
+    }
+
+    resolutionHandler(context, winner) {
+        this.game.addMessage('{0} wins the duel and will not bow as a result of this conflict\'s resolution', winner);
+        this.game.actions.cardLastingEffect({ effect: AbilityDsl.effects.doesNotBow() }).resolve(winner, context);
+    }
+}
+
+HonorableChallenger.id = 'honorable-challenger';
+
+module.exports = HonorableChallenger;
--- a/server/game/GameActions/LastingEffectAction.ts
+++ b/server/game/GameActions/LastingEffectAction.ts
@@ -7,7 +7,7 @@ export interface LastingEffectGeneralProperties extends GameActionProperties {
     duration?: Durations;
     condition?: (context: AbilityContext) => boolean;
     until?: WhenType;
-    effect: any[];
+    effect: any;
 }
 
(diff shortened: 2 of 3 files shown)


server/game/GameActions/LastingEffectCardAction.ts
```diff
@@ -5,6 +5,8 @@ import BaseCard = require('../basecard');
 import { LastingEffectGeneralProperties } from './LastingEffectAction';
 
 export interface LastingEffectCardProperties extends LastingEffectGeneralProperties {
+    match?: (card: BaseCard, context?: AbilityContext) => boolean;
```
jeremylarner: Neither of these properties should be necessary - the match is always the target of the game action, and match should always be a card, not a function for a lasting effect, so targetController is unnecessary

server/game/cards/06-CotE/HonorableChallenger.js
```diff
@@ -0,0 +1,39 @@
...
+const { Players, CardTypes } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
+const GameActions = require('../../GameActions/GameActions');
+
+class HonorableChallenger extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Initiate a military duel',
+            condition: () => this.isParticipating(),
+            target: {
```
jeremylarner: This could maybe use the new `initiateDuel` property?

server/game/cards/06-CotE/HonorableChallenger.js
```diff
@@ -0,0 +1,39 @@
...
+                    resolutionHandler: (winner) => this.resolutionHandler(context, winner)
+                }))
+            }
+        });
+    }
+
+    resolutionHandler(context, winner) {
+        this.game.addMessage('{0} wins the duel and will not bow as a result of this conflict\'s resolution', winner);
+
+        let doesNotBowEffect = GameActions.cardLastingEffect({
```
jeremylarner: As you're already importing AbilityDsl, I'd rather you used `AbilityDsl.actions.cardLastingEffect` here

server/game/cards/06-CotE/HonorableChallenger.js
```diff
@@ -0,0 +1,39 @@
...
+            match: card => card === winner,
+            targetController: Players.Any,
+            effect: [AbilityDsl.effects.doesNotBow()]
```
jeremylarner: You shouldn't need `match` or `targetController` here.  You also don't need the effect to be an array, but that's style and readability rather than being objectively wrong

server/game/cards/06-CotE/HonorableChallenger.js
```diff
@@ -0,0 +1,39 @@
...
+        }).getEvent(context.source, context);
```
jeremylarner: I think I'd prefer you to use `.resolve(winner, context)` here.  `cardLastingEffect` should always target the card which is getting the lasting effect - the game action can tell where the effect is coming from because of the context object

Issue #2708, opened 2018-12-22
CotE - Infiltrator's Tools


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/50371725-7ebdcc80-05b8-11e9-9cfe-9dc28d93ed85.png)

SArendeli:
I'll code this one up.

PR #2739, opened 2019-01-03
CotE - Infiltrator's Tools

SArendeli:
Not sure if it's necessary (or generally considered redundant), but I added tests for this card even though it's a pretty straightforward attachment.


jamesoddy:
Closes #2708

PR #2756, opened 2019-01-08
CotE - Loyal Challenger

### LINKED ISSUE #2736: CotE - Loyal Challenger
image

jamesoddy:
Closes #2736 

I needed to add in `messageArgs` for `DelayedEffect` in order to display a decent message.  Please check that this has been done correctly / is needed.


jeremylarner:
What you've done looks fine, but I'm not really happy with the way we do messaging in general.  I would like to find a better solution across all the instances which use it

Issue #2713, opened 2018-12-29
CotE - Kiku Matsuri


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/50534580-76284180-0b36-11e9-80ee-8d43582b7ec4.png)

SArendeli:
I'll take this

PR #2757, opened 2019-01-09
CotE - Kiku Matsuri

SArendeli:
Closes Issue #2713


server/game/cards/06-CotE/KikuMatsuri.js
```diff
@@ -0,0 +1,28 @@
+const ProvinceCard = require('../../provincecard.js');
+const AbilityDsl = require('../../abilitydsl');
+
+class KikuMatsuri extends ProvinceCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Honor a character home from each side',
+            targets: {
+                myCharacter: {
+                    cardCondition: (card, context) => card.isParticipating() && card.controller === context.player,
```
jeremylarner: For both of these it would be better to have `cardType: CardTypes.Character` and `controller: Players.Self/Opponent`.  You should always have a cardType in a target, because it makes the prompt work correctly, and using controller (rather than adding the condition to the cardCondition) will make the targeting prompt work better

SArendeli: No problem, I'll fix that up shortly

PR #2768, opened 2019-01-11
Fix for rename of Akodo Toturi 2 (again!)

jamesoddy:
Build will still fail because of changes for Mitsu/KI fix.

This should change akodo-toturi-neutral back to akodo-toturi-2


jeremylarner:
Included in #2770

PR #2776, opened 2019-01-14
CotE: Unmatched Expertise

Kwaice:


server/game/cards/06-CotE/UnmatchedExpertise.js
```diff
@@ -0,0 +1,21 @@
...
+        this.whileAttached({
+            effect: ability.effects.cardCannot('becomeDishonored')
+        });
+        this.forcedReaction({
+            title: 'Removed after attached character loses a conflict',
+            when: {
+                afterConflict: (event, context) => context.source.parent.isParticipating() &&
+                                                   event.conflict.loser === context.source.parent.controller
+            },
+            gameAction: ability.actions.discardFromPlay(context => ({ target: context.source }))
```
jeremylarner: You shouldn't need the `context => ({ target: context.source })` here, as it should default to context.source (and not passing it a function will make the code run faster)

PR #2781, opened 2019-01-16
Fix for Reclusive Zokojin covert

### LINKED ISSUE #2661: Reclusive Zokujin's conditional covert
Opponent declares an attack with a guy with covert.  Picks Fire Ring.  Target's Zokujin with covert.  So far, so good.
Then, the opponent changes their mind and decides to attack at Earth instead.  The Zokujin remains coverted, even though he is no longer a legal target for it.

This can also play out in the other direction where you attack with Zokujin on earth, covert a guy, and then switch the ring to something else before finalizing the attack.  The target's border remains purple.

jamesoddy:
Closes #2661 

Includes commit 9aaeede (`passConflilct()` helper function)

`recalculateCovert()`, that is called after each click in declare attackers, will now:

* remove any coverted defenders that have somehow now gained covert, and
* remove any coverted defenders until the number of attackers with covert equals (or is fewer than) the number of coverted defenders

I also added a few tests for Reclusive Zokujin to test for this situation and also had to slightly change the `initiateconflictprompt.spec` to mock the `isCovert()` function.

### DIFF
--- a/server/game/gamesteps/conflict/initiateconflictprompt.js
+++ b/server/game/gamesteps/conflict/initiateconflictprompt.js
@@ -178,7 +178,7 @@ class InitiateConflictPrompt extends UiPrompt {
 
     recalculateCovert() {
         // remove any selected defenders with covert
-        this.selectedDefenders.map(card => {
+        _.each(this.selectedDefenders, card => {
             if(card.isCovert()) {
                 card.covert = false;
                 this.selectedDefenders.splice(this.selectedDefenders.indexOf(card));
--- a/server/game/gamesteps/conflict/initiateconflictprompt.js
+++ b/server/game/gamesteps/conflict/initiateconflictprompt.js
@@ -93,7 +93,6 @@ class InitiateConflictPrompt extends UiPrompt {
         }
 
         return this.selectCard(card);
-
     }
 
     onRingClicked(player, ring) {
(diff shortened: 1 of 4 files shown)


server/game/gamesteps/conflict/initiateconflictprompt.js
```diff
@@ -178,7 +177,21 @@ class InitiateConflictPrompt extends UiPrompt {
     }
 
     recalculateCovert() {
+        // remove any selected defenders with covert
+        this.selectedDefenders.map(card => {
```
jeremylarner: I don't really like using `.map` to iterate over a function....

jamesoddy: This is my ignorance of JS array functions.

What is best to use `_.each`?

jeremylarner: You could use `_.each` or a for loop

server/game/gamesteps/conflict/initiateconflictprompt.js
```diff
@@ -178,7 +177,21 @@ class InitiateConflictPrompt extends UiPrompt {
     }
 
     recalculateCovert() {
+        // remove any selected defenders with covert
+        _.each(this.selectedDefenders, card => {
```
jeremylarner: Maybe I'm not understanding what this is supposed to do, but wouldn't `.filter` achieve what you're aiming to do here?

jamesoddy: Probably?  I need to set `.covert = false` on the card and also remove the card from the `selectedDefenders` array.  There may be a more concise way of doing this - any ideas?

jeremylarner:
I'd like to spend some more time looking at this before we merge it - probably after we release CotE

PR #2817, opened 2019-01-24
CotE - Prudent Challenger

### LINKED ISSUE #2787: CotE - Prudent Challenger

jamesoddy:
Closes #2787

### DIFF
--- a/server/game/cards/06-CotE/PrudentChallenger.js
+++ b/server/game/cards/06-CotE/PrudentChallenger.js
@@ -1,4 +1,5 @@
 const DrawCard = require('../../drawcard.js');
+const { CardTypes } = require('../../Constants');
 
 class PrudentChallenger extends DrawCard {
     setupCardAbilities() {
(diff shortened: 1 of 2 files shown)


server/game/cards/06-CotE/PrudentChallenger.js
```diff
@@ -0,0 +1,36 @@
...
+                this.game.promptWithHandlerMenu(context.player, {
+                    activePromptTitle: 'Choose an attachment to discard',
+                    context: context,
+                    cards: loser.attachments.sortBy(card => card.name),
+                    cardHandler: card => {
+                        this.game.addMessage('{0} chooses to discard {1}', context.player, card);
+                        this.game.applyGameAction(context, { discardCard: card });
+                    }
+                });
+            }
```
jeremylarner: a) I think we should be using a SelectCardPrompt, not a HandlerMenuPrompt here
b) This is a perfect example of the kind of card which is going to cause major headaches with Hantei - I suspect that we need to restructure duels to have a gameAction rather than a resolutionHandler....

jamesoddy: a) I have changed to a select prompt as requested
b) Yes a winnerGameAction and a loserGameAction would be ideal and while we are at it I think we could do with standardising the chat messages for duels, maybe something like `Duel Result - Winner: Prudent Challenger, Loser: Doji Whisperer`

PR #2820, opened 2019-01-26
unveilded destiny implementation

### LINKED ISSUE #2804: CotE - Unveiled Destiny
image

Matthew-Griffith:
closes #2804 

the keeper/seeker role and the RoleCard classes were changed to include a method to determine the element that need to be added to the contest for the this attachment.

### DIFF
--- a/server/game/Effects/EffectBuilder.js
+++ b/server/game/Effects/EffectBuilder.js
@@ -30,7 +30,7 @@ const EffectBuilder = {
     ring: {
         static: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new StaticEffect(type, value))),
         dynamic: (type, values) => _.map(values, (value) => {
-            (game, source, props) => new RingEffect(game, source, props, new DynamicEffect(type, value));
+            return (game, source, props) => new RingEffect(game, source, props, new DynamicEffect(type, value));
         }),
         detached: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new DetachedEffect(type, value.apply, value.unapply))),
         flexible: (type, value) => _.isFunction(value) ? EffectBuilder.ring.dynamic(type, value) : EffectBuilder.ring.static(type, value)
--- a/server/game/Effects/EffectBuilder.js
+++ b/server/game/Effects/EffectBuilder.js
@@ -29,13 +29,7 @@ const EffectBuilder = {
     },
     ring: {
         static: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new StaticEffect(type, value))),
-        dynamic: (type, values) => {
-            let effectsArray = [];
-            for(let i = 0; i < values.length; i++) {
-                (game, source, props) => effectsArray.push(new RingEffect(game, source, props, new DynamicEffect(type, values[i])));
-            }
-            return effectsArray;
-        },
+        dynamic: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new DynamicEffect(type, value))),
         detached: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new DetachedEffect(type, value.apply, value.unapply))),
         flexible: (type, value) => _.isFunction(value) ? EffectBuilder.ring.dynamic(type, value) : EffectBuilder.ring.static(type, value)
     }
--- a/server/game/Effects/EffectBuilder.js
+++ b/server/game/Effects/EffectBuilder.js
@@ -29,9 +29,14 @@ const EffectBuilder = {
     },
     ring: {
         static: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new StaticEffect(type, value))),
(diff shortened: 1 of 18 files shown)


server/game/cards/06-CotE/UnveiledDestiny.js
```diff
@@ -0,0 +1,16 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl.js');
+
+class UnveiledDestiny extends DrawCard {
+    setupCardAbilities() {
+        this.persistentEffect({
+            condition: context => context.source.parent.isAttacking() && context.player.role.getElement() !== [],
```
jeremylarner: I don't think this will work due to the way js deals with Arrays.  `getElement()` will return a reference (or pointer) to an Array, and `[]` will be a reference to a different Array, so this `!==` will always evaluate as `true` even when both arrays are empty.  However, due to the way you've done the `effect` below, I think you can just remove this altogether. You should check that the player does actually have a role card here though, so I would put `condition: context => context.player.role && context.source.parent.isAttacking()`

server/game/cards/06-CotE/UnveiledDestiny.js
```diff
@@ -0,0 +1,16 @@
...
+            match: ring => ring.contested,
+            effect: context => context.player.role.getElement().forEach(element => AbilityDsl.effects.addElement(element))
```
jeremylarner: I think `map` is what you mean here.... this should be `context => context.player.role.getElement().map(element => AbilityDsl.effects.addElement(element))`

jeremylarner: Looking at this again, it's more tricky that I first thought.  You can't pass a function to `effect`.  There are two ways to implement this I think:
1) Create 5 persistent effects, one for each element, which condition on whether that element is on the controllers role card. The problem with doing it this way is that it's only a single effect on the physical card
2) Change the way the `addElement` effect works.  It needs to be changed to be `flexible` instead of `static` and then changed so that it takes an array of elements as well as a single element.  Then you can pass the effect a function which returns `getElement()`

If you want to go with option 2, but you'd prefer not to delve into the guts of the engine, let me know and I'll do that bit

Matthew-Griffith: I'll definitely give option 2 a shot. Thanks for the feedback

Matthew-Griffith: so with my latest commit I think getElement should be able to an array of elements. so now that its flexible it will run `EffectBuilder.ring.dynamic` I changed that function so it will map the array given so it will return an array of `RingEffect` classes instead of just one. 

then the `presistentEffect` function on the `UnveilingDestiny` class will add this array of `RingEffect` classes to the engine via the `addEffectToEngine` method which is defined of the `EffectSource` class. this method already handles arrays of effects as well as single effect, so I think it should work.

I am going to start writing tests for this, let me know if I have an incorrect understand of the working of the engine.

server/game/Effects/EffectBuilder.js
```diff
@@ -29,7 +29,13 @@ const EffectBuilder = {
...
     ring: {
         static: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new StaticEffect(type, value))),
-        dynamic: (type, value) => ((game, source, props) => new RingEffect(game, source, props, new DynamicEffect(type, value))),
+        dynamic: (type, values) => {
+            let effectsArray = [];
+            for(let i = 0; i < values.length; i++) {
+                (game, source, props) => effectsArray.push(new RingEffect(game, source, props, new DynamicEffect(type, values[i])));
+            }
+            return effectsArray;
+        },
```
jeremylarner: I'm not sure what you're trying to do here, but I'm pretty sure this won't do it :)
I don't think you need any change in EffectBuilder.js at all.  The change you need to make to make this work is in ring.js: Change `return _.uniq(this.getEffects(EffectNames.AddElement).concat([this.element]));` to `return _.uniq(_.flatten(this.getEffects(EffectNames.AddElement).concat([this.element])));`

server/game/cards/06-CotE/UnveiledDestiny.js
```diff
@@ -0,0 +1,16 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl.js');
+
+class UnveiledDestiny extends DrawCard {
+    setupCardAbilities() {
+        this.persistentEffect({
+            condition: context => context.player.role && context.source.parent.isAttacking(),
+            match: ring => ring.contested,
+            effect: AbilityDsl.effects.addElement(context => context.player.role.getElement())
```
jeremylarner: The `calculate` function for dynamic effects gets passed the target of the effect first (the contested ring in this case), so you need `(ring, context) => context.player.role.getElement()` here

Issue #2750, opened 2019-01-06
CotE - Warm Welcome


jeremylarner:
![image](https://user-images.githubusercontent.com/31585532/50740842-b3eb3100-11ec-11e9-84be-44e5174d84d8.png)

SArendeli:
I can take this one

jeremylarner:
This looks like Kyuden Isawa or Togashi Mitsu, but it's actually templated differently.  I think it needs two game actions (playCard and moveCard), not a post handler

jeremylarner:
Basically, you want something like Togashi Mitsu, but instead of including a `postHandler`, you want a separate `moveCard` game action (so `gameAction` should be an array with `playCard()` and `moveCard()` in it)

SArendeli:
Okay sounds good, I'll take a shot at it

PR #2821, opened 2019-01-26
CotE - Warm Welcome

SArendeli:
Implementation and Test for #2750 

Help Needed, can't get the sequence of playCard / moveCard to work in combination


jeremylarner:
Yes, sorry that's probably my fault - I sent you off on a bit of a wild goose chase, didn't I...
I think putting your `gameAction` array in a `AbilityDsl.actions.sequential()` should maybe fix it though?

jeremylarner:
I changed this to use sequentialAction - seems to pass all tests now

PR #2822, opened 2019-01-30
Cote - Duels

### LINKED ISSUE #2615: CotE : Mirumoto Hitomi

### LINKED ISSUE #2748: CotE - True Strike Kenjutsu
image

jamesoddy:
Closes #2748 
Closes #2615 

Changes include:

* Duel honor bid no longer changes statistic (just adds to the result of the duel)
* Added `DuelTypes` enum
* Duels can now target multiple characters (for Mirumoto Hitomi)
* Renamed functions/variables to include statistic instead of skill where appropriate (to macth RRG description)
* Added the ability to specify a player for `ChooseGameAction`
* Added `BaseMilitary` and `BasePolitical` duel types (for True Strike Kenjutsu)
* Implementation and tests for Hitomi and Kenjutsu


jeremylarner:
I still don't like having BaseMil and BasePol duel types, I think it would be better to have an alternative skill calc function.

I think the solution to the hitomi problem is to make chooseAction deal better with multiple targets - it needs to prompt for a choice for each one.

My thinking is definitely moving in the direction of having a gameAction property on duels, and getting rid of resolution handlers

jamesoddy:
That's absolutely fine.  I did start doing it that way and reconsidered when I thought about the chat message that would be shown.  I'll work it out though.

Ok I will take a look at the chooseAction multiple targeting.

Ok if we leave moving to gameActions on duels as a separate thing?  I don't fancy making this PR any bigger if I can avoid it.

PR #2823, opened 2019-01-31
CotE - Righteous magistrate

### LINKED ISSUE #2803: CotE - Righteous Magistrate
image

Matthew-Griffith:
closes #2803 

First implementation of Righteous Magistrate with a few test to make sure, players can't player cards with honor cost, don't lose/gain honor from honored/dishonored character leaving play and can't lose honor from card reactions.

### DIFF
--- a/server/game/cards/06-CotE/RighteousMagistrate.js
+++ b/server/game/cards/06-CotE/RighteousMagistrate.js
@@ -1,0 +1,27 @@
+const DrawCard = require('../../drawcard.js');
+const { Players } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl.js');
+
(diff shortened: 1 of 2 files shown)


jeremylarner:
This looks great! Would you mind adding a test to check duels work correctly while she is defending? I think they should do everything like normal, but no honor should be transfered if one player bids higher than the other

jeremylarner:
Sorry, I didn't mean to close the PR - fat fingers.... :(

jeremylarner:
Could you also add a test to check that neither Assassination or Smuggling Deal can be played?

Matthew-Griffith:
I don't think you can play assassination since you can pay the cost of the card.

PR #2825, opened 2019-02-01
CotE - Mirumoto Dojo

### LINKED ISSUE #2788: CotE - Mirumoto Dôjô

jamesoddy:
Closes #2788 

Includes PR #2822 (for the DuelTypes enum)

### DIFF
--- a/server/game/CardAbility.js
+++ b/server/game/CardAbility.js
@@ -9,16 +9,17 @@ class CardAbility extends ThenAbility {
         if(properties.initiateDuel) {
             if(properties.condition) {
                 let condition = properties.condition;
-                properties.condition = context => context.source.isParticipating() && condition(context);
+                properties.condition = context => (context.source.type !== CardTypes.Character || context.source.isParticipating()) && condition(context);
             } else {
-                properties.condition = context => context.source.isParticipating();
+                properties.condition = context => context.source.type !== CardTypes.Character || context.source.isParticipating();
             }
             properties.targets = {
                 challenger: {
                     cardType: CardTypes.Character,
-                    mode: TargetModes.AutoSingle,
+                    mode: card.type === CardTypes.Character ? TargetModes.AutoSingle : TargetModes.Single,
                     controller: Players.Self,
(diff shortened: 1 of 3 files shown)


jeremylarner:
In future, I think it might be better not to mix PRs like this.  I realise not having the previous one causes tests to break, but for review purposes, it makes it really difficult to see what's going on.  I'm happy to be overruled if others disagree though

jamesoddy:
I wass hoping the idea would be to approve and merge in PR 2788 first then this one should just simplify to the implementation and tests for Mirumoto Dojo.  I take your point though - I'll try and avoid it in future.

jeremylarner:
Yeah, I still have 31 files changed :(

jamesoddy:
Yeah, realised it wasn't going to work because the duels PR is going to be squashed when it is merged so the commit history is completely different.  I've sorted it out and now this is just the changes needed for Mirumoto Dojo.

jeremylarner:
Ah, it's my fault for squashing? I'd never made the connection.... Sorry :/

server/game/CardAbility.js
```diff
@@ -9,16 +9,17 @@ class CardAbility extends ThenAbility {
         if(properties.initiateDuel) {
             if(properties.condition) {
                 let condition = properties.condition;
-                properties.condition = context => context.source.isParticipating() && condition(context);
+                properties.condition = context => (context.source.type !== CardTypes.Character || context.source.isParticipating()) && condition(context);
```
jeremylarner: Given the target conditions, do we need these extra condition checks? (I know that's not something you added, just pointing it out)

jamesoddy:
Haha, not blaming you for squashing :) .  I think squashing is best to clean up the commits.  I just hadn't considered it in this case that it means that including other PRs in a PR will never clean up once the other PR is merged.

server/game/CardAbility.js
jamesoddy: Yeah I think you are right.  I'll push a new commit.

PR #2828, opened 2019-02-02
CotE - Seppun hidden guard

### LINKED ISSUE #2791: CotE - Seppun Hidden Guard
shg

Matthew-Griffith:
closes #2791 

here is the early version of the card class. it looks like it mostly working expect the random discard part of the card.

### DIFF
--- a/server/game/cards/06-CotE/SeppunHiddenGuard.js
+++ b/server/game/cards/06-CotE/SeppunHiddenGuard.js
@@ -16,8 +16,7 @@ class SeppunHiddenGuard extends DrawCard {
             effectArgs: context => context.event.card,
             handler: context => {
                 context.cancel();
-                // AbilityDsl.actions.discardAtRandom();
-                context.game.applyGameAction(context, { discardAtRandom: context.player.opponent });
+                context.game.applyGameAction(context, { discardAtRandom: context.event.context.player });
             }
         });
     }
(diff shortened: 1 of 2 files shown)


server/game/cards/06-CotE/SeppunHiddenGuard.js
```diff
@@ -0,0 +1,28 @@
...
+                    card.isUnique() && card.controller === context.player && card.location === Locations.PlayArea)
+                )
+            },
+            cost: AbilityDsl.costs.sacrificeSelf(),
+            effect: 'cancel the effects of {1}',
+            effectArgs: context => context.event.card,
+            handler: context => {
+                context.cancel();
+                // AbilityDsl.actions.discardAtRandom();
+                context.game.applyGameAction(context, { discardAtRandom: context.player.opponent });
```
jeremylarner: In theory, you could use this ability to cancel your own ability, in which case you should discard a card.  I think this should target `context.event.context.player`

Matthew-Griffith: good catch, I didn't think about that when I read the card

jeremylarner:
By default, the client doesn't prompt you to cancel your own abilities.  You'll need to set that option for the test player if you want to test that functionality

Matthew-Griffith:
oh thanks, I was really confused why that test was failing

PR #2830, opened 2019-02-02
Rework Mantis Tenkinja

jeremylarner:


server/game/costs/ReduceableFateCost.js
```diff
@@ -1,102 +1,40 @@
...
-                        properties.numberOfChoices = properties.maxFate - properties.minFate + 1;
-                        if(result.cancelled || properties.numberOfChoices === 0) {
-                            return;
-                        }
-                        this.promptForAlternateFate(context, result, properties);
-                    });
-                }
+        if(this.getReducedCost(context) > maxPlayerFate + alternatePoolTotal) {
+            result.cancelled = true;
+        } else if(!result.cancelled && alternatePools.length > 0 && context.player.checkRestrictions('takeFateFromRings', context)) {
```
jamesoddy: Seems a bit strange to me that you would be checking for 'takeFateFromRings' restrictions here.  Correct me if I'm wrong but I would have thought the idea was to be able to have different alternate fate pools that aren't necessarily from the fate rings.

jeremylarner: Yeah, I kind of left it open so that non-ring fate pools fit easily in in the future.  However, for the moment, they don't exist, and it didn't seem worth doing extra work to distinguish between until that changes

jamesoddy: Sure.  Just something to watch out for in the future.  I could certainly see them printing a holding that builds up fate or something.  Actually, that sounds like a cool idea for a card :)

PR #2833, opened 2019-02-05
CotE - Defend Your Honor

### LINKED ISSUE #2805: CotE - Defend Your Honor
image

jamesoddy:
Closes #2805

### DIFF
--- a/server/game/cards/06-CotE/DefendYourHonor.js
+++ b/server/game/cards/06-CotE/DefendYourHonor.js
@@ -1,5 +1,5 @@
 const DrawCard = require('../../drawcard.js');
-const { CardTypes, DuelTypes, Players } = require('../../Constants');
+const { CardTypes, DuelTypes } = require('../../Constants');
 
 class DefendYourHonor extends DrawCard {
     setupCardAbilities() {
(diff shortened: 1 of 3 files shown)


server/game/cards/06-CotE/DefendYourHonor.js
```diff
@@ -0,0 +1,49 @@
...
+                duelTarget: {
+                    dependsOn: 'challenger',
+                    cardType: CardTypes.Character,
+                    controller: Players.Opponent,
+                    cardCondition: card => card.isParticipating()
+                }
+            },
+            effect: 'initiate a military duel between {1} and {2}',
+            effectArgs: (context) => [context.targets.challenger, context.targets.duelTarget],
+            handler: context => {
```
jeremylarner: Why is this a handler not a duel game action?

jamesoddy: Short answer is it didn't work as a game action.  I didn't look too deep into it, but `wouldInterrupt` would not trigger when I used a gameAction at all.  It looks like all the `wouldInterrupt`s only use `handler` so I assume that it is not setup to handle `gameAction`.

Any help to get that working would be much appreciated.

jamesoddy:
I'm currently assuming that you choose the challenger before your opponent chooses the duel target as that is probably RAI if not quite RAW at the moment.

PR #2834, opened 2019-02-06
DeckSearchAction improvements/fixes

jamesoddy:
This PR is doing a few things.

* ~~Extending the `DeckSearchAction` to be a bit more generic (allowing you to override the `cardHandler`, `takeNothingHandler`,~~ default to reveal if there is a `cardCondition`, ~~and allow you to specify the conflict deck or dynasty deck)~~.  This simultaneously fixes some cards that would not reveal when they should.
* Explicitly set reveal on Tattered Missive
* Additional matcher for testing purposes (`toHaveDisabledPromptButton`) and adjustment to `toHavePromptButton` and `clickPrompt` to deal with disabled prompt buttons
* Integration tests for Agasha Swordsmith

### DIFF
--- a/server/game/GameActions/DeckSearchAction.ts
+++ b/server/game/GameActions/DeckSearchAction.ts
@@ -15,14 +15,15 @@ export class DeckSearchAction extends PlayerAction {
     eventName = EventNames.OnDeckSearch;
 
     defaultProperties: DeckSearchProperties = {
-        amount: -1,
-        reveal: false,
-        cardCondition: () => true
+        amount: -1
     };
 
     getProperties(context: AbilityContext, additionalProperties = {}): DeckSearchProperties {
         let properties = super.getProperties(context, additionalProperties) as DeckSearchProperties;
-        properties.reveal = properties.reveal || properties.cardCondition.toString() !== this.defaultProperties.cardCondition.toString();
+        if(properties.reveal === undefined) {
+            properties.reveal = properties.cardCondition !== undefined;            
+        }
+        properties.cardCondition = properties.cardCondition || (() => true);
         return properties;
     }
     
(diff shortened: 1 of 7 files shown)


jeremylarner:
I think using this with Miya Library in its current form is problematic because it always shuffles the deck, doesn't it?

server/game/GameActions/DeckSearchAction.ts
```diff
@@ -20,6 +20,12 @@ export class DeckSearchAction extends PlayerAction {
         cardCondition: () => true
     };
 
+    getProperties(context: AbilityContext, additionalProperties = {}): DeckSearchProperties {
+        let properties = super.getProperties(context, additionalProperties) as DeckSearchProperties;
+        properties.reveal = properties.reveal || properties.cardCondition.toString() !== this.defaultProperties.cardCondition.toString();
```
jeremylarner: This seems really ugly - what are you trying to do here?

jamesoddy: This is to

if the `reveal:` property has not been set then set `reveal:` to be `true` if `cardCondition:` is not the default (i.e. it has been set) otherwise it will be `false`

The idea is if there is a card condition then you would have to reveal it, but you can still override the implementation by explicitly setting `reveal:`.

The ugly bit is checking whether `cardCondition:` has been set - there may be a better way of doing that?

jeremylarner: It will override a `reveal: false` - is that what you meant to do? If you want to check whether cardCondition has been set then I think you should not set it in defaultProperties....

jamesoddy: Yes you are right.  I didn't particularly want to not set cardCondition to () => true as  a default otherwise you have to check it in multiple places to see if it is defined - that seems more messy to me.  I've pushed a new version that hopefully makes the best of both worlds.

PR #2856, opened 2019-02-08
Change currentDuel to work with nested duels

jeremylarner:


server/game/Duel.js
```diff
@@ -32,6 +33,10 @@ class Duel {
         return (card === this.challenger || card === this.target || this.target.includes(card)) && card.location === Locations.PlayArea;
     }
 
+    isInvolvedInAnyDuel(card) {
+        return this.isInvolved(card) || this.previousDuel && this.previousDuel.isInvolved(card);
+    }
```
jamesoddy: This isn't going to work in a nest of 3 or more duels right?  I think you want:
```
return this.isInvolved(card) || this.previousDuel && this.previousDuel.isInvolvedInAnyDuel(card);
```
to recursively look up the tree.

jeremylarner: Yes, you're right - fixed

PR #2860, opened 2019-02-09
CotE - Leniency

### LINKED ISSUE #2852: CotE - Leniency
image

Matthew-Griffith:
closes #2852

here is how I thought the card would work but the `gameAction: AbilityDsl.actions.putIntoPlay()`
seems to not put the character into play


jeremylarner:
You can't have `handler:` and `gameAction:` in the same card. handler will override any game actions.  So either we need to create a cancel game action or you need to put the putIntoPlay in the handler.

jeremylarner:
After thinking about this for a bit, I think we should have a cancel/replace game action - I'll write one and submit a PR

jeremylarner:
PR for a cancel action is #2864 
You want to add this as a `gameAction` in `target` with `putIntoPlay` as `replacementGameAction`

Matthew-Griffith:
Thanks @jeremylarner. I'll add those changes tonight

jeremylarner:
PR is merged, so you should be good to go

Matthew-Griffith:
for some reason the test for defend the wall is failing. it lets the player do the reaction of leniency but they can't select the border rider for some reason.

PR #2867, opened 2019-02-09
CotE: Unmask

### LINKED ISSUE #2850: CotE - Unmask
image

CDMayberry:
Closes #2850

### DIFF
--- a/server/game/cards/06-CotE/Unmask.js
+++ b/server/game/cards/06-CotE/Unmask.js
@@ -1,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const { Players, CardTypes } = require('../../Constants');
+
+class Unmask extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Discard a character\'s status token and set its skills to its printed value until the conflicts end. Controller gains 2 honor.',
+            condition: context => context.player.opponent && context.player.showBid > context.player.opponent.showBid,
+            target: {
+                cardType: CardTypes.Character,
+                controller: Players.Any,
+                cardCondition: card => card.isParticipating(),
+                gameAction: [
+                    ability.actions.discardStatusToken(),
+                    ability.actions.cardLastingEffect(context => ({
+                        effect: [
+                            ability.effects.setBaseMilitarySkill(context.target.getBaseMilitarySkill()),
+                            ability.effects.setBasePoliticalSkill(context.target.getBasePoliticalSkill())
+                        ]
+                    })),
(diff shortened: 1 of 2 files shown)


server/game/cards/06-CotE/Unmask.js
```diff
@@ -0,0 +1,30 @@
...
+            target: {
+                cardType: CardTypes.Character,
+                controller: Players.Any,
+                cardCondition: card => card.isParticipating(),
+                gameAction: ability.actions.multiple([
+                    ability.actions.discardStatusToken(),
+                    ability.actions.cardLastingEffect(context => ({
+                        effect: [
+                            ability.effects.setMilitarySkill(context.target.getBaseMilitarySkill()),
+                            ability.effects.setPoliticalSkill(context.target.getBasePoliticalSkill())
```
jamesoddy: Setting to base isn't necessarily the same as setting to their printed values because of cards like Way of the Lion.  I think you can use:
```js
ability.effects.setMilitarySkill(context.target.printedMilitarySkill),
ability.effects.setPoliticalSkill(context.target.printedPoliticalSkill)
```
instead.

server/game/cards/06-CotE/Unmask.js
```diff
@@ -0,0 +1,30 @@
...
+                        ]
+                    })),
+                    ability.actions.gainHonor(context => ({ amount: 2, target: context.target.controller }))
```
jeremylarner: Technically the gainHonor action should be outside the target object.
Also, you need an effect property

server/game/cards/06-CotE/Unmask.js
CDMayberry: Ah that's right. Updated.

server/game/cards/06-CotE/Unmask.js
CDMayberry: Moved the honor gain and added the effect message.

server/game/cards/06-CotE/Unmask.js
```diff
@@ -0,0 +1,32 @@
+const DrawCard = require('../../drawcard.js');
+const { Players, CardTypes } = require('../../Constants');
+
+class Unmask extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Discard a character\'s status token and set its skills to its printed value until the conflicts end. Its controller gains 2 honor.',
```
jeremylarner: This is way too long - remember that the reason for a title is to allow a player to distinguish between abilities in a prompt.

CDMayberry: Changed to "Discard a character's status token and set skills to printed value"

PR #2878, opened 2019-02-12
CotE -Severed From the Stream

### LINKED ISSUE #2849: CotE - Severed from the Stream
image

jamesoddy:
Closes #2849 

Please note that I added a new function `getGloryCount()` to player and referenced that in both this implementation and `conflictphase.js` (`claimImperialFavor`).

Also, I am assuming that this cannot be triggered if glory counts are equal.  This may change in the future if anything triggers off "Perform a glory count".


jeremylarner:
I think this can be triggered if the glory counts are equal.... I'm not sure though

PR #2881, opened 2019-02-13
New Conditional meta game action

jamesoddy:
Expects parameters:

`gameAction:`
`condition?:`
`else:`

Will resolve gameAction: if condition: is true or will resolve else: if condition: is not true.  If condition: is not defined then will resolve else: if gameAction: cannot effect game state.

### DIFF
--- a/server/game/GameActions/ConditionalGameAction.ts
+++ b/server/game/GameActions/ConditionalGameAction.ts
@@ -1,0 +1,66 @@
+import { GameAction, GameActionProperties } from './GameAction';
+import AbilityContext = require('../AbilityContext');
+
+export interface ConditionalActionProperties extends GameActionProperties {
+    gameAction: GameAction;
+    condition: boolean | ((context: AbilityContext) => boolean);
+    else?: GameAction;
+}
+
+export class ConditionalGameAction extends GameAction {
+    effect = 'do a conditional action';
+
+    constructor(properties: ConditionalActionProperties | ((context: AbilityContext) => ConditionalActionProperties)) {
+        super(properties);
+    }
+
+    getProperties(context: AbilityContext, additionalProperties = {}): ConditionalActionProperties {
+        let properties = super.getProperties(context, additionalProperties) as ConditionalActionProperties;
+        properties.condition = properties.condition || true;
+        return properties;
+    }
+
+    meetsCondition(context: AbilityContext, properties: ConditionalActionProperties): boolean {
+        if(typeof properties.condition === 'function') {
+            return properties.condition(context);
+        }
+        return properties.condition as boolean;
+    }
+
+    hasLegalTarget(context: AbilityContext, additionalProperties = {}): boolean {
+        let properties = this.getProperties(context, additionalProperties);
+        if(this.meetsCondition(context, properties)) {
+            return properties.gameAction.hasLegalTarget(context, additionalProperties) || properties.else.hasLegalTarget(context, additionalProperties);
+        }
+        if(properties.else) {
+            return properties.else.hasLegalTarget(context, additionalProperties);
+        }
+        return false;
+    }
+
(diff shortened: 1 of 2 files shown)


server/game/GameActions/ConditionalGameAction.ts
```diff
@@ -0,0 +1,66 @@
...
+import AbilityContext = require('../AbilityContext');
+
+export interface ConditionalActionProperties extends GameActionProperties {
+    gameAction: GameAction;
+    condition: boolean | ((context: AbilityContext) => boolean);
+    else?: GameAction;
+}
+
+export class ConditionalGameAction extends GameAction {
+    effect = 'do a conditional action';
```
jeremylarner: I think we should probably give this an effect message which tells the player what it's going to do

server/game/GameActions/ConditionalGameAction.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    constructor(properties: ConditionalActionProperties | ((context: AbilityContext) => ConditionalActionProperties)) {
+        super(properties);
```
jeremylarner: The constructor here shouldn't be needed?

server/game/GameActions/ConditionalGameAction.ts
```diff
@@ -0,0 +1,66 @@
+import { GameAction, GameActionProperties } from './GameAction';
+import AbilityContext = require('../AbilityContext');
+
+export interface ConditionalActionProperties extends GameActionProperties {
+    gameAction: GameAction;
+    condition: boolean | ((context: AbilityContext) => boolean);
```
jeremylarner: Would this ever be boolean?  I think it should probably always be a function....

server/game/GameActions/ConditionalGameAction.ts
```diff
@@ -0,0 +1,66 @@
...
+    effect = 'do a conditional action';
+
+    constructor(properties: ConditionalActionProperties | ((context: AbilityContext) => ConditionalActionProperties)) {
+        super(properties);
+    }
+
+    getProperties(context: AbilityContext, additionalProperties = {}): ConditionalActionProperties {
+        let properties = super.getProperties(context, additionalProperties) as ConditionalActionProperties;
+        properties.condition = properties.condition || true;
+        return properties;
```
jeremylarner: This should probably be handled with defaultProperties.  You should put something in here though which sets the two game action default targets to be this actions target

server/game/GameActions/ConditionalGameAction.ts
```diff
@@ -0,0 +1,66 @@
...
+        if(typeof properties.condition === 'function') {
+            return properties.condition(context);
+        }
+        return properties.condition as boolean;
+    }
+
+    hasLegalTarget(context: AbilityContext, additionalProperties = {}): boolean {
+        let properties = this.getProperties(context, additionalProperties);
+        if(this.meetsCondition(context, properties)) {
+            return properties.gameAction.hasLegalTarget(context, additionalProperties) || properties.else.hasLegalTarget(context, additionalProperties);
```
jeremylarner: This will throw an error if gameAction isn't legal and properties.else isn't defined

jeremylarner:
The only thing I think is potentially confusing here is that if the condition is true but gameAction is illegal, what happens?  I can see cases where you want else to happen, but also cases where you don't....

jamesoddy:
Potential cards that could use this
```
Deathseeker, Mirumoto Raitsugu, Calling in Favors, Karada District
Akodo makoto
CiF and Karada are a bit of a mess atm
so you might need to be a bit creative to get them to work
Also the new dragon holding might be a good fit for a conditional action
```

jeremylarner:
This needs to be PRed to develop (or somewhere else)

PR #2940, opened 2019-02-19
Challenge on the Fields - nested duels

jamesoddy:
Just tests at the moment.

### DIFF
--- a/server/game/AbilityContext.ts
+++ b/server/game/AbilityContext.ts
@@ -17,6 +17,7 @@ interface AbilityContextProperties {
     events?: any[];
     stage?: Stages;
     targetAbility?: any;
+    duel?: any;
 }
 
 class AbilityContext {
@@ -31,6 +32,7 @@ class AbilityContext {
     events: any[] = [];
     stage: Stages;
     targetAbility: any;
+    duel: any;
     target: any;
     select: string;
     ring: Ring;
@@ -48,6 +50,7 @@ class AbilityContext {
         this.selects = properties.selects || {};
         this.stage = properties.stage || Stages.Effect;
         this.targetAbility = properties.targetAbility;
+        this.duel = properties.duel;
     }
 
     copy(newProps: object): AbilityContext {
@@ -88,7 +91,8 @@ class AbilityContext {
             selects: Object.assign({}, this.selects),
             events: this.events,
             stage: this.stage,
-            targetAbility: this.targetAbility
+            targetAbility: this.targetAbility,
+            duel: this.duel
         };
     }
 }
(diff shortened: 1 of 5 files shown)


server/game/AbilityContext.ts
```diff
@@ -17,6 +17,7 @@ interface AbilityContextProperties {
     events?: any[];
     stage?: Stages;
     targetAbility?: any;
+    duel?: any;
```
jeremylarner: I don't think we should need duel on the context object... Why is this necessary?

server/game/GameActions/DuelAction.ts
```diff
@@ -102,9 +103,17 @@ export class DuelAction extends CardGameAction {
             context.game.addMessage('The duel cannot proceed as at least one participant for each side has to be in play');
             return;
         }
+        var duel = new Duel(context.game, properties.challenger, cards, properties.type, properties.statistic);
+        if(properties.duringDuelAction) {
+            properties.duringDuelAction.duration = Durations.UntilEndOfDuel;
+            context.duel = duel;
```
jeremylarner: Ok, I think I get what you're trying to do, but I don't think it will work - does this fix pass tests?  The context of the ability which initiated the duel is not the same as the context of the lasting effect once it's created.  I can see an argument that it perhaps should be, but I think that will require a fairly major rewrite to the effect engine.

server/game/GameActions/DuelAction.ts
```diff
@@ -102,9 +103,17 @@ export class DuelAction extends CardGameAction {
             context.game.addMessage('The duel cannot proceed as at least one participant for each side has to be in play');
             return;
         }
+        var duel = new Duel(context.game, properties.challenger, cards, properties.type, properties.statistic);
+        if(properties.duringDuelAction) {
+            properties.duringDuelAction.duration = Durations.UntilEndOfDuel;
```
jeremylarner: Game actions are stateless now, so you cannot change the duration like this (I don't think).  You would have to change either the properties or the propertyFactory property of the game action, and that get's very messy.  I don't think there's a way to make it work without introducing state to game actions, which I don't want to do (having just done so much work to remove it)

jeremylarner:
I think I would try an alternative approach.  If you look at effects with custom durations, you can pass an `until` property which works similar to a `when` property in handlers (it returns a boolean for the specified event).  I think it will be easier to do an `challengerEffect` and `targetEffect` rather than a general any-game-action-is-fair-game approach.  I think the general approach would be great, but I'm not sure there's a good way to do to it elegantly

jeremylarner:
Tests included in #3059

PR #2967, opened 2019-02-23
Solitary Strength fix

Kwaice:
Fix for Solitary Strength being discarded if attached to an opposing character


jeremylarner:
Do we have a ruling saying that the card checks whether its controller has any other characters in the conflict?  The card text says 'with 1 or more other characters you control'....

Kwaice:
Mmmm, you're right. Unless Tyler clarifies that, let's not merge it i guess.

PR #2974, opened 2019-02-26
Bugfix for Hantei Peasant's Advice interaction

jeremylarner:


gryffon:
Can you grab the changes from develop, just to make sure builds pass?

PR #2975, opened 2019-02-26
WotW - Shinomen wayfinders

### LINKED ISSUE #2956: WotW: Shinomen Wayfinders

Matthew-Griffith:
closes #2956 

here is the first version of the card with one test but the test is throwing an error that it can't find the card.

### DIFF
--- a/server/game/cards/07-WotW/ShinomenWayfinders.js
+++ b/server/game/cards/07-WotW/ShinomenWayfinders.js
@@ -8,7 +8,7 @@ class ShinomenWayfinders extends DrawCard {
             location: Locations.Any,
             targetController: Players.Any,
             effect: AbilityDsl.effects.reduceCost({
-                amount: (player) => {
+                amount: (card, player) => {
                     return player.filterCardsInPlay(card => {
                         return card.isParticipating() && card.isFaction('unicorn');
                     }).length;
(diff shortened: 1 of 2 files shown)


server/game/cards/07-WotW/ShinomenWayfinders.js
```diff
@@ -0,0 +1,25 @@
...
+        this.persistentEffect({
+            location: Locations.PlayArea,
+            targetController: Players.Self,
+            effect: AbilityDsl.effects.reduceCost({
+                playingTypes: PlayTypes.PlayFromHand,
+                amount: context => {
+                    return context.player.filterCardsInPlay(card => {
+                        return card.isParticipating() && card.isFaction('unicorn');
+                    }).length;
+                },
```
jeremylarner: A couple of problems here:
- I think the cost reduction should work from anywhere
- It should function for both players (if I use infiltrator to play your Wayfinders, I get the reduction too)
- I think it should effect both PlayTypes (I should get the reduction if I play it from BAH - I guess that doesn't really matter, but you never know what they will print in the future)
- I think amount can be a function, but it takes the card being played as it's parameter, not the context object.  That might be problematic in the case where it's played using Infiltrator...

jeremylarner:
I think the tests are failing because we haven't updated the develop branch with the latest spoilers from 5rdb

server/game/cards/07-WotW/ShinomenWayfinders.js
Matthew-Griffith: so I added the things you asked for expect I made the context the only parameter to the function for the amount. I am not really sure how you would determine the number of attacking characters without having the context

PR #2985, opened 2019-03-02
WotW - Ring of binding

### LINKED ISSUE #2895: WotW - Ring of Binding
image

Matthew-Griffith:
closes #2895 

waiting to finish testing once the card is add to the repo

### DIFF
--- a/server/game/cards/07-WotW/RingOfBinding.js
+++ b/server/game/cards/07-WotW/RingOfBinding.js
@@ -5,7 +5,7 @@ const AbilityDsl = require('../../abilitydsl');
 class RingOfBinding extends DrawCard {
     setupCardAbilities() {
         this.persistentEffect({
-            match: (context) => context.game.currentPhase === Phases.Fate && context.game.getPlayersInFirstPlayerOrder()[0] === context.player,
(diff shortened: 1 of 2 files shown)


server/game/cards/07-WotW/RingOfBinding.js
```diff
@@ -0,0 +1,19 @@
+const DrawCard = require('../../drawcard.js');
+const { Phases } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
+
+class RingOfBinding extends DrawCard {
+    setupCardAbilities() {
+        this.persistentEffect({
+            match: (context) => context.game.currentPhase === Phases.Fate && context.game.getPlayersInFirstPlayerOrder()[0] === context.player,
```
jeremylarner: I think this should be a `condition:` not a `match:` Also `context.player.firstPlayer` is a more elegant way of checking if context.player is the first player :P

Matthew-Griffith: I made the changes you suggested but the first two tests are still failing. do you have any ideas what the bug might be.

PR #2987, opened 2019-03-02
WotW - Ikoma Anakazu

### LINKED ISSUE #2977: WotW: Ikoma Anakazu
image0-1

jamesoddy:
Closes #2977

### DIFF
--- a/server/game/cards/07-WotW/IkomaAnakazu.js
+++ b/server/game/cards/07-WotW/IkomaAnakazu.js
@@ -1,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const EventRegistrar = require('../../eventregistrar.js');
+const AbilityDsl = require('../../abilitydsl');
+const { EventNames } = require('../../Constants');
+
+class IkomaAnakazu extends DrawCard {
+    setupCardAbilities() {
+        this.opponentBrokenProvinceThisPhase = false;
+        this.eventRegistrar = new EventRegistrar(this.game, this);
+        this.eventRegistrar.register([EventNames.OnBreakProvince, EventNames.OnPhaseEnded]);
+
+        this.persistentEffect({
+            condition: context => context.source.isParticipating() && this.opponentBrokenProvinceThisPhase,
+            effect: AbilityDsl.effects.modifyBothSkills(3)
+        });
+    }
+
+    onPhaseEnded() {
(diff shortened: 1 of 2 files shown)


server/game/cards/07-WotW/IkomaAnakazu.js
```diff
@@ -0,0 +1,31 @@
...
+            effect: AbilityDsl.effects.modifyBothSkills(3)
+        });
+    }
+
+    onPhaseEnded() {
+        this.opponentBrokenProvinceThisPhase = false;
+    }
+
+    onBreakProvince(event) {
+        if(event.conflict && event.conflict.attackingPlayer !== this.controller) {
```
jeremylarner: This won't work if his ability is copied by an opposing e.g. Yokuni.  Look at e.g. Fire Elemental Guard for some ideas

jeremylarner:
On second look it's a constant ability, not a triggered ability, so Yokuni can't copy.  And it can't be actressed or Miyako'd, or taken control of.  It's a conflict character though, so it will break (I think) if it gets played with Sand Road or Infiltrator by your opponent

jamesoddy:
It's a good catch and probably a good idea to take account of it if just for a potential future card.  I will take a look.

I love that you are always on the lookout for Yokuni interactions - maybe you need some therapy for the PTSD from implementing him :P

jeremylarner:
Definitely... he was so hard.... I think it was mostly because I had no idea what I was doing back then, and I just hacked everything :/

server/game/cards/07-WotW/IkomaAnakazu.js
```diff
@@ -0,0 +1,35 @@
...
+const { EventNames } = require('../../Constants');
+
+class IkomaAnakazu extends DrawCard {
+    setupCardAbilities() {
+        this.brokenProvincesThisPhase = {};
+        this.eventRegistrar = new EventRegistrar(this.game, this);
+        this.eventRegistrar.register([EventNames.OnBreakProvince, EventNames.OnPhaseEnded]);
+
+        this.persistentEffect({
+            condition: context => context.source.isParticipating() && this.brokenProvincesThisPhase[context.source.controller.opponent.name] > 0,
```
jeremylarner: This will throw an error in single player games.  Also, you probably don't need to track the number of provinces, just having a boolean is fine. I guess it doesn't really matter though

PR #2991, opened 2019-03-04
WotW - Agasha Taiko

### LINKED ISSUE #2978: WotW: Agasha Taiko
image0-2

jamesoddy:
Closes #2978 

Created a new effect `cannotBeAttacked` for this implementation.


jeremylarner:
How does this interact with Chisei District and the other one I can't remember the name of?  They are cannot be declared instead, which is different?

jamesoddy:
This should be independent from the cannot initiate conflict effects we already have.  It is my understanding that they are separate effects as this effect will prevent Talisman moving the conflict to the affected province (as well as preventing conflict initiation).

Issue #2971, opened 2019-02-25
Chisei District


DojiHoe:
Chisei District is not active.  It does not stop an opponent from declaring a military conflict on the province it is located on.

jamesoddy:
Thanks for the report - I've managed to replicate your issue.  This only happens if you change to political **after** you have chosen the province to attack.

The declaration of conflicts is slightly broken as it will allow you to change selections of conflict type and ring after you have selected the province and chosen covert targets etc.  This is actually causing a problem with a number of cards (Zokojin for example).  The whole declaration needs a bit of a re-work that we will get to at some point.  Thanks for the report.

jamesoddy:
Conflict declaration has been refactored and is now live (PR #3135) that should fix this issue.

PR #2992, opened 2019-03-04
Tests for Chisei District

jamesoddy:
Relates to #2971

This is currently only a test (that fails).

The `initateConflictPrompt` needs a refactor to check all conditions between all state changes and remove invalid selections.

Also relates to PR #2781


jeremylarner:
We need to fix Fire Tensai Acolyte as well.... are we waiting on a ruling there?

jeremylarner:
These tests are now in #3135

PR #2994, opened 2019-03-04
WotW - Visiting Advisor

### LINKED ISSUE #2957: WotW: Visiting Advisor

jamesoddy:
Closes #2957

### DIFF
--- a/server/game/cards/07-WotW/VisitingAdvisor.js
+++ b/server/game/cards/07-WotW/VisitingAdvisor.js
@@ -12,11 +12,9 @@ class VisitingAdvisor extends DrawCard {
                 cardType: CardTypes.Character,
                 optional: true,
                 cardCondition: (card, context) => card !== context.source,
-                gameAction: AbilityDsl.actions.multiple([
-                    AbilityDsl.actions.sendHome(),
-                    AbilityDsl.actions.sendHome(context => ({ target: context.source }))
-                ])
+                gameAction: AbilityDsl.actions.sendHome()
             },
+            gameAction: AbilityDsl.actions.sendHome(context => ({ target: context.source })),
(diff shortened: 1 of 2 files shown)


jeremylarner:
Does this work if Visiting Advisor is the only character in the conflict?

jamesoddy:
It did work if Visiting Advisor was your only participating character (you just have to click the prompt for "No more targets").  I've added a test for that situation.

I have updated as you suggested though - my only concern with that was that the actions were not "combined" and there could be some weird reaction interaction.  But there is nothing that I can think of off the top of my head that would cause an issue.

jeremylarner:
The tests for this will break when combined with #3013.  The implementation may break too :P

server/game/cards/07-WotW/VisitingAdvisor.js
```diff
@@ -0,0 +1,26 @@
...
+            target: {
+                controller: Players.Self,
+                cardType: CardTypes.Character,
+                optional: true,
+                cardCondition: (card, context) => card !== context.source,
+                gameAction: AbilityDsl.actions.sendHome()
+            },
+            gameAction: AbilityDsl.actions.sendHome(context => ({ target: context.source })),
+            effect: 'send {0}{1}{2} home',
+            effectArgs: (context) => context.target.length === 0 ? [context.source] : [' and ', context.source]
```
jeremylarner: This looks awkward to me.  Do you need an `effect` at all here?  Doesn't the default just work(TM)?

jamesoddy: Damn right it was awkward, but as it is also affecting `context.source` (Visiting Advisor itself) this was the only way I could get it to display the right messages.

PR #3008, opened 2019-03-12
Fix lobby order, minor tweaks. 

Syrail:
Lobby game list is no longer reversed.
Removed extraneous ring state from game summary object (ring state is part of game state still).
Replaced [private] with unicode &#128274; symbol for private games.

### DIFF
--- a/server/game/game.js
+++ b/server/game/game.js
@@ -1232,13 +1232,6 @@ class Game extends EventEmitter {
             name: this.name,
             owner: _.omit(this.owner, ['blocklist', 'email', 'emailHash', 'promptedActionWindows', 'settings']),
             players: playerSummaries,
-            rings: {
-                air: this.rings.air.getState(activePlayer),
-                earth: this.rings.earth.getState(activePlayer),
-                fire: this.rings.fire.getState(activePlayer),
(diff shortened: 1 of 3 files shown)


server/game/game.js
```diff
@@ -1232,13 +1232,6 @@ class Game extends EventEmitter {
             name: this.name,
             owner: _.omit(this.owner, ['blocklist', 'email', 'emailHash', 'promptedActionWindows', 'settings']),
             players: playerSummaries,
-            rings: {
-                air: this.rings.air.getState(activePlayer),
-                earth: this.rings.earth.getState(activePlayer),
-                fire: this.rings.fire.getState(activePlayer),
-                void: this.rings.void.getState(activePlayer),
-                water: this.rings.water.getState(activePlayer)
-            },
```
jamesoddy: What was this change for?

jamesoddy: The build is failing on lint because the activePlayer is no longer a used variable.

jamesoddy: Ah sorry, just saw in your description why it was removed.  It still needs fixing so that `activePlayer` is removed as it is no longer a used variable.

PR #3013, opened 2019-03-15
WotW - Warriors of the Wind

### LINKED ISSUE #2958: WotW: Warriors of the Wind

jamesoddy:
Closes #2958

### DIFF
--- a/server/game/cards/07-WotW/WarriorsOfTheWind.js
+++ b/server/game/cards/07-WotW/WarriorsOfTheWind.js
@@ -1,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl');
+const { CardTypes, Players } = require('../../Constants');
+
+class WarriorsOfTheWind extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Re-arrange participating cavalry characters',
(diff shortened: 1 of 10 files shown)


server/game/cards/07-WotW/WarriorsOfTheWind.js
```diff
@@ -0,0 +1,31 @@
...
+            title: 'Re-arrange participating cavalry characters',
+            condition: () => this.game.isDuringConflict(),
+            gameAction: AbilityDsl.actions.sendHome(context => ({
+                target: context.player.filterCardsInPlay(card => card.hasTrait('cavalry') && card.isParticipating())
+            })),
+            then: {
+                target: {
+                    activePromptTitle: 'Choose characters',
+                    numCards: 0,
+                    //optional: true,
```
jamesoddy: As soon as this is marked as optional then the "then" part never triggers at all, but it needs to be possible to select 0 characters to move into the conflict.

@jeremylarner It would be great if you could take a look.

jeremylarner: #3015 seems to fix this

jamesoddy:
This includes commits from PR #3015 
Now set to use `mode: TargetModes.Unlimited` instead of `numCards: 0`
and added a test for Hantei interaction

PR #3026, opened 2019-04-03
WotW: Wild Stallion

### LINKED ISSUE #3023: WotW: Wild Stallion
image

Kwaice:
Closes #3023


jamesoddy:
Ok, I'm going to be pedantic here - you've misspelled it "Stalion" throughout instead of "Stallion" (double "l").  This will actually cause a problem when we get the fiveringsdb data as (hopefully) they would give it the id wild-stallion.

Actually, I would love it if we called it "wyld-stallyns" - any objections?

test/server/cards/07-WotW/WildStallion.spec.js
```diff
@@ -0,0 +1,74 @@
...
+            });
+
+            it('should let you chose no characters to move to the conflict', function() {
+                this.initiateConflict({
+                    attackers: [this.motoYouth],
+                    defenders: []
+                });
+                this.player2.pass();
+                this.player1.clickCard(this.wildStallion);
+                this.player1.clickPrompt('No More Targets');
```
jamesoddy: This was recently changed in PR #3015 to require 'Done' instead of 'No more targets'
```
travis error
 Error: Couldn't click on "No More Targets" for player1. Current prompt is:
    Choose a character
    [ Done ]
    [ Pay costs first ]
    [ Cancel ]
```

PR #3066, opened 2019-04-24
MotC - Asahina Takamori

### LINKED ISSUE #2897: MotC - Asahina Takamori
image

jamesoddy:
Closes #2897

### DIFF
--- a/server/game/cards/08-MotC/AsahinaTakamori.js
+++ b/server/game/cards/08-MotC/AsahinaTakamori.js
@@ -1,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const { CardTypes, Durations, Players } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
+
+class AsahinaTakamori extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            title: 'Choose a character to pacify',
+            when: {
+                onCardPlayed: (event) => event.card.type === CardTypes.Character && event.card.isFaction('crane')
+            },
+            target: {
+                controller: Players.Opponent,
+                cardType: CardTypes.Character,
+                cardCondition: (card, context) => card.getCost() <= context.event.card.getCost(),
+                gameAction: AbilityDsl.actions.cardLastingEffect({
+                    duration: Durations.UntilEndOfRound,
+                    effect: [
+                        AbilityDsl.effects.cannotParticipateAsAttacker(),
+                        AbilityDsl.effects.cannotParticipateAsDefender()
+                    ]
+                })
+            },
+            effect: 'prevent {0} from being declared as an attacker or defender this round'
+        });
+    }
+}
+
+AsahinaTakamori.id = 'asahina-takamori';
+
+module.exports = AsahinaTakamori;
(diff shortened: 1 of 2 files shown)


server/game/cards/08-MotC/AsahinaTakamori.js
```diff
@@ -0,0 +1,31 @@
...
+            },
+            target: {
+                controller: Players.Opponent,
+                cardType: CardTypes.Character,
+                cardCondition: (card, context) => card.getCost() <= context.event.card.getCost(),
+                gameAction: AbilityDsl.actions.cardLastingEffect({
+                    duration: Durations.UntilEndOfRound,
+                    effect: [
+                        AbilityDsl.effects.cannotParticipateAsAttacker(),
+                        AbilityDsl.effects.cannotParticipateAsDefender()
```
Kwaice: I think it should be declareAsAttacker here.

jamesoddy: Ah good call.  Fix incoming.

jamesoddy: Huh, I thought we had that as an effect already, but doesn't look like it.  I'll have to write a new one :)

jamesoddy: Nevermind, found it as a `cardCannot` effect

server/game/cards/08-MotC/AsahinaTakamori.js
```diff
@@ -0,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const { CardTypes, Durations, Players } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
+
+class AsahinaTakamori extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            title: 'Choose a character to pacify',
```
jeremylarner: This is a title, not a prompt, so I would make it 'Pacify a character'

server/game/cards/08-MotC/AsahinaTakamori.js
```diff
@@ -0,0 +1,31 @@
...
+            when: {
+                onCardPlayed: (event) => event.card.type === CardTypes.Character && event.card.isFaction('crane')
```
jeremylarner: You need to check that `context.player` is playing the card

server/game/cards/08-MotC/AsahinaTakamori.js
```diff
@@ -0,0 +1,31 @@
...
+            },
+            target: {
+                controller: Players.Opponent,
+                cardType: CardTypes.Character,
+                cardCondition: (card, context) => card.getCost() <= context.event.card.getCost(),
```
jeremylarner: You should use `card.costLessThan(num)` to deal with cards with no cost being immune to these effects

server/game/cards/08-MotC/AsahinaTakamori.js
jamesoddy: Good catch.  I've added a test and fix for that.

server/game/cards/08-MotC/AsahinaTakamori.js
jamesoddy: Changed.

gryffon:
Looks like this has some test failures.

PR #3074, opened 2019-04-24
Motc: Kakita Favorite

### LINKED ISSUE #2902: MotC - Kakita Favorite
image

siowyisheng:
closes #2902 

first card implementation. tests are passing locally.

### DIFF
--- a/server/game/cards/08-MotC/KakitaFavorite.js
+++ b/server/game/cards/08-MotC/KakitaFavorite.js
@@ -1,0 +1,16 @@
+const DrawCard = require('../../drawcard.js');
+
+class KakitaFavorite extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: context =>
+                this.game.currentDuel &&
+                this.game.currentDuel.isInvolvedInAnyDuel(context.source),
+            effect: ability.effects.modifyPoliticalSkill(2)
+        });
+    }
+}
+
+KakitaFavorite.id = 'kakita-favorite';
+
+module.exports = KakitaFavorite;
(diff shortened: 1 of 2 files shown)


server/game/cards/08-MotC/KakitaFavorite.js
```diff
@@ -0,0 +1,16 @@
+const DrawCard = require('../../drawcard.js');
+
+class KakitaFavorite extends DrawCard {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            condition: () =>
+                this.game.currentDuel &&
+                this.game.currentDuel.isInvolvedInAnyDuel(this),
```
jeremylarner: As a general rule, you should avoid using `this` in abilities, because it will cause problems if an effect copies this cards ability text.  `condition` gets `context`, so here what you should have is `condition: context => this.game.currentDuel && this.game.currentDuel.isInvolvedInAnyDuel(context.source)`

test/server/cards/08-MotC/KakitaFavorite.spec.js
```diff
@@ -0,0 +1,46 @@
...
+                this.borderRider = this.player1.findCardByName('border-rider');
+
+                this.noMoreActions();
+                this.initiateConflict({
+                    attackers: ['border-rider'],
+                    defenders: ['kakita-favorite']
+                });
+            });
+
+            it('should not have extra political outside of a conflict', function() {
```
jeremylarner: conflict should be duel here?

siowyisheng: yep my bad

server/game/cards/08-MotC/KakitaFavorite.js
siowyisheng: got it!

siowyisheng:
hm... the typescript errors continue to be there, but the jasmine tests are otherwise passing. not sure what's going on.

siowyisheng:
the requested changes have been made btw

jeremylarner:
It's the package you've added to package.json which is causing the problem.  Your card and test work fine without it.  Can you remove that change?

siowyisheng:
done

PR #3076, opened 2019-04-24
Add conditional game action

jeremylarner:


jeremylarner:
Right.  I'm going to add an 'if able' action to cover the other case

PR #3093, opened 2019-04-26
Tests - Shinjo Ambusher - Akodo Toshiro interaction

### LINKED ISSUE #3042: unexpected interaction between Akodo Toshiro and Shinjo Ambusher
I have tested with my friend. If I use Akodo Toshiro's action ability first, then my Shinjo Ambusher cannot trigger the reaction ability. However, if I use the reaction ability of Shinjo Ambusher first, the province will break even I trigger Akodo Toshiro's action ability after.

jamesoddy:
Closes #3042 

Kind of just putting this here so I don't forget about it - just failing tests at the moment.


jeremylarner:
I looked at this, but it's not failing tests for me

jamesoddy:
I've rebased and now tests are passing so we must have inadvertently fixed the issue somewhere else :)

PR #3097, opened 2019-04-26
MotC: The Spear Rushes Forth

### LINKED ISSUE #2906: MotC - The Spear Rushes Forth
image

Kwaice:
Closes #2906 

I think it works like it's supposed to but cost.dishonor might not be the right one.

Couldn't find a cost.discardStatusToken

### DIFF
--- a/server/game/cards/08-MotC/TheSpearRushesForth.js
+++ b/server/game/cards/08-MotC/TheSpearRushesForth.js
@@ -6,7 +6,7 @@ class TheSpearRushesForth extends DrawCard {
     setupCardAbilities() {
         this.action({
             title: 'Bow a participating character',
-            condition: () => this.game.isDuringConflict(),
+            condition: () => this.game.isDuringConflict('military'),
             cost: AbilityDsl.costs.dishonor({
                 cardCondition: card => card.isHonored && card.isParticipating()
             }),
--- a/server/game/cards/08-MotC/TheSpearRushesForth.js
+++ b/server/game/cards/08-MotC/TheSpearRushesForth.js
@@ -1,0 +1,27 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl');
+const { CardTypes, Players } = require('../../Constants');
+
+class TheSpearRushesForth extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Bow a participating character',
+            condition: () => this.game.isDuringConflict(),
+            cost: AbilityDsl.costs.dishonor({
+                cardType: CardTypes.Character,
+                controller: Players.Self,
+                cardCondition: card => card.isHonored && card.isParticipating()
+            }),
+            target: {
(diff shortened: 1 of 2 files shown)


server/game/cards/08-MotC/TheSpearRushesForth.js
```diff
@@ -0,0 +1,27 @@
...
+
+class TheSpearRushesForth extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Bow a participating character',
+            condition: () => this.game.isDuringConflict(),
+            cost: AbilityDsl.costs.dishonor({
+                cardType: CardTypes.Character,
+                controller: Players.Self,
+                cardCondition: card => card.isHonored && card.isParticipating()
```
jeremylarner: this is fine - you only need `cardCondition`, because costs only allow you to pick your own characters by default, and the game knows that only characters can be dishonored

Kwaice: Makes sense. Thanks !

server/game/cards/08-MotC/TheSpearRushesForth.js
```diff
@@ -0,0 +1,25 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl');
+const { CardTypes, Players } = require('../../Constants');
+
+class TheSpearRushesForth extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Bow a participating character',
+            condition: () => this.game.isDuringConflict(),
```
jamesoddy: Needs to be only during a military conflict

Kwaice: I totally missed that, thanks.

test/server/cards/08-MotC/TheSpearRushesForth.spec.js
```diff
@@ -0,0 +1,104 @@
...
+                this.player1.clickCard(this.theSpearRushesForth);
+                expect(this.player1).toBeAbleToSelect(this.dojiChallenger);
+                expect(this.player1).toBeAbleToSelect(this.kakitaYoshi);
+                this.player1.clickCard(this.kakitaYoshi);
+                this.player1.clickCard(this.kakitaKaezin);
+                expect(this.kakitaYoshi.bowed).toBe(true);
+                expect(this.kakitaKaezin.isHonored).toBe(false);
+            });
+
+            it('should work even if steward of law is present in the conflict', function() {
```
jamesoddy: Steward of law wouldn't prevent going from honored to ordinary anyway, a better test would be with Young Harrier which would normally prevent dishonoring (but not the removal of an honor status token).

Kwaice: I added the Young Harrier test and it indeed prevents the cost from being paid in this case.

jeremylarner:
Included in #3133

PR #3104, opened 2019-04-28
MotC - Maker of Keepsakes

### LINKED ISSUE #3044: MotC - Maker of Keepsakes
image

jamesoddy:
Closes #3044 

I'm pretty sure this is functionally the same as "cannot become dishonored", but let me know if there is some scenario that I need to check for.


jeremylarner:
I'm pretty sure I'm going to change the name of the effect to `receiveDishonorTokens` when I rework the status token stuff, so we'll need to change the effect then

PR #3105, opened 2019-04-29
 MotC - Bayushi Dairu

### LINKED ISSUE #3050: MotC - Bayushi Dairu
image

jeremylarner:
Closes #3050


server/game/GameActions/DiscardStatusAction.ts
```diff
@@ -1,26 +1,15 @@
...
-        if(card.location !== Locations.PlayArea || !card.isHonored && !card.isDishonored) {
-            return false;
-        }
-        return super.canAffect(card, context);
-    }
 
     eventHandler(event): void {
-        event.card.isHonored = false;
-        event.card.isDishonored = false;
+        event.token.card.makeOrdinary();
```
jamesoddy: This will (I think) still set as ordinary even if the token has changed, right?  Unless we set card as null when the token is removed from play (see comment below)

server/game/GameActions/MoveTokenAction.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    addPropertiesToEvent(event, token: StatusToken, context: AbilityContext, additionalProperties = {}): void {
+        const { recipient } = this.getProperties(context) as MoveTokenProperties;
+        super.addPropertiesToEvent(event, token, context, additionalProperties);
+        event.recipient = recipient;
+    }
+
+    eventHandler(event): void {
+        event.token.card.makeOrdinary();
```
jamesoddy: Again, do we need to check that the token has not been removed before resolution?

server/game/drawcard.js
```diff
@@ -354,22 +353,38 @@ class DrawCard extends BaseCard {
...
+            this.makeOrdinary();
         } else {
-            this.isDishonored = true;
+            this.setPersonalHonor(new StatusToken(this.game, this, false));
         }
     }
 
+    makeOrdinary() {
+        this.setPersonalHonor();
+    }
```
jamesoddy: I think it may be worth setting `card` to `null` on any personalHonor that may already be present on the card so that if the token changes before resolution then the `.token.card` would be `null`.

server/game/Events/InitiateCardAbilityEvent.js
```diff
@@ -9,10 +9,12 @@ class InitiateCardAbilityEvent extends Event {
...
             this.ringTargets = _.flatten(_.values(this.context.rings));
             this.selectTargets = _.flatten(_.values(this.context.selects));
+            this.tokenTargets = _.flatten(_.values(this.context.tokens));
         } else {
             this.cardTargets = [];
             this.ringTargets = [];
             this.selectTargets = [];
+            this.tokenTargets = [];
         }
         this.allTargets = this.cardTargets.concat(this.ringTargets.concat(this.selectTargets));
```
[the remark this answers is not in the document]
jeremylarner: Yup, fixed ty :)

server/game/GameActions/DiscardStatusAction.ts
jeremylarner: Yes, you're right - see solution below

server/game/GameActions/MoveTokenAction.ts
jeremylarner: Yes :)

server/game/drawcard.js
jeremylarner: I don't really want to set `card` to `null` as it loses information.  I've added tests which check that the token in the event is still the token on the card, which I think achieves the same aim.

PR #3113, opened 2019-04-30
Change meta-action properties

jeremylarner:
This change replaces actionParameter with a more flexible additionalProperties function


jeremylarner:
I really don't like `additionalProperties` as a property name - any alternative name suggestions gratefully received!

PR #3115, opened 2019-04-30
Isawa Tadaka RRG v9 errata

jeremylarner:


jeremylarner:
> We should add a test validating that Tadaka is in effect while the Earth ring is contested.

There is already a test for this...?

PR #3117, opened 2019-04-30
TotSR bug

Matthew-Griffith:
these test show you can't play your opponent's "action:" cards off TotSR


jeremylarner:
Contained in #3118

PR #3125, opened 2019-05-01
MotC - Courteous Scheming

### LINKED ISSUE #3053: MotC - Courteous Scheming
image

siowyisheng:
closes #3053

### DIFF
--- a/server/game/cards/08-MotC/CourteousScheming.js
+++ b/server/game/cards/08-MotC/CourteousScheming.js
@@ -1,0 +1,31 @@
+const DrawCard = require('../../drawcard.js');
+const { Durations, DuelTypes, Players } = require('../../Constants');
+
+class CourteousScheming extends DrawCard {
+    setupCardAbilities(ability) {
+        this.action({
+            title: 'Initiate a political duel',
+            condition: () =>
+                this.game.currentConflict &&
+                this.game.currentConflict.conflictType === 'political',
+            initiateDuel: context => ({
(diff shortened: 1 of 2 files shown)


server/game/cards/08-MotC/CourteousScheming.js
```diff
@@ -0,0 +1,28 @@
...
+            condition: () =>
+                this.game.currentConflict &&
+                this.game.currentConflict.conflictType === 'political',
+            initiateDuel: () => ({
+                type: DuelTypes.Political,
+                opponentChoosesDuelTarget: true,
+                gameAction: duel =>
+                    duel.winner &&
+                    ability.actions.playerLastingEffect({
+                        targetController: duel.winner.controller,
```
jeremylarner: This should be a Players constant I think, so I'm surprised this is passing tests, as it shouldn't work on the opponent like this!

siowyisheng: thanks! i think Courtly Challenger is also doing

```
gameAction: duel => AbilityDsl.actions.draw({
    amount: 2,
    target: duel.winner.controller
})
```

jeremylarner: Courtly challenger isn't applying a lasting effect.  Player lasting effects work in an unintuitive way due to legacy code in the effect engine.  I should get round to refactoring it to work the way everyone expects it to work, but yeah... time

PR #3126, opened 2019-05-01
MotC - Distinguished Dojo

### LINKED ISSUE #2899: MotC - Distinguished Dojo
image

jamesoddy:
Closes #2899 

Includes #3121 Add AddTokenAction

Not sure if UI is working - needs checking locally (wasn't working for me but then neither was Public Forum).

Also fixes a few duels that weren't playing nice with draws.


jamesoddy:
Got the subThenContext to work not sure what I was doing wrong previously. :)

Issue #3040, opened 2019-04-14
Loyal Challenger interaction when "stolen" via Shosuro Actress Bug


ruudoniac:
Loyal Challenger's passive ability triggers as if my opponent controls it , although i have her under my control via Shosuro Actress (ex I win the conflict with her and he loses honor instead of me getting honor), guess same stands for Blackmail , not tested that though.

jamesoddy:
Confirmed and fixed in PR #3154

Thanks for the report!

PR #3146, opened 2019-05-04
Loyal challenger bug

Matthew-Griffith:
tests for issue #3040

### DIFF
--- a/test/server/cards/06-CotE/LoyalChallenger.spec.js
+++ b/test/server/cards/06-CotE/LoyalChallenger.spec.js
@@ -1,18 +1,22 @@
-describe('Loyal Challenger', function() {
+fdescribe('Loyal Challenger', function() {
     integration(function() {
         describe('Loyal Challenger', function() {
             beforeEach(function() {
                 this.setupTest({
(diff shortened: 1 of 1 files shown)


Matthew-Griffith:
I just realized the test is wrong the first player gains two off the air ring. I'll fix it in the morning

jamesoddy:
@Matthew-Griffith I've created a PR to your branch which has a test which I suspect was the problem and a fix.

jamesoddy:
Sorry, I should have explained further.  Loyal Challenger was looking at `context.player` rather than `context.source.controller`, which I suspect (not 100% sure) will be the player than first plays Loyal Challenger - your tests are passing because Loyal Challenger was already in the discard.  The new test gets player1 to play Loyal Challenger first, then assasinated and then Shosuro Actress takes control.  That fails unless we change to `context.source.controller`.

Matthew-Griffith:
that's great I just wrote the test off the issue reported by a player and it didn't have a screen shot. so I just wrote a basic test and saw it was passing and did pay it a ton of attention.

PR #3147, opened 2019-05-04
tests for bug with shosuro hyobu

Matthew-Griffith:
currently the card is triggering when a unique character is discard for the dupe rule. and is not triggering when spies at courts is played.


Matthew-Griffith:
she is also triggering when the controller discards cards from UA

gryffon:
Closing in preparation of migration. If necessary, can be re-opened against new location.

PR #3148, opened 2019-05-04
iuchi farseer bug with charge

Matthew-Griffith:
adds a test to show charge currently is not working with iuchi farseer


jeremylarner:
Can you have a go at fixing this?  I think it's just a question of changing the when trigger

Matthew-Griffith:
yeah I'll try fixing it tonight

PR #3153, opened 2019-05-06
Refactor ring select

The-Outcast:
- moved Ring.jsx render responsibility to parent component.
- removed Ring.jsx class calculation to hide itself.
- removed unnecessary small declaration on ring childitems while this can be managed by the container.
- Rings will now appear in the middle if they are relevant for being selectable.


gryffon:
Closing in preparation of migration. Can be re-opened against new repository if needed.

PR #3158, opened 2019-05-07
Initiate conflict bugfixes

jeremylarner:
- also add a cost message for discarding status token

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -282,11 +282,17 @@ class Player extends GameObject {
         if(conflictRing.length === 0) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


server/game/player.js
```diff
@@ -964,7 +970,7 @@ class Player extends GameObject {
 
         let location = card.location;
 
-        if(location === Locations.PlayArea || (card.type === CardTypes.Holding && provinceLocations.includes(location))) {
+        if(location === Locations.PlayArea || (card.type === CardTypes.Holding && provinceLocations.includes(location) && !provinceLocations.includes(targetLocation))) {
```
jamesoddy: Was this meant to be part of the Zentaro PR rather than here?

PR #3178, opened 2019-05-09
BoB - Duel To The Death

### LINKED ISSUE #3171: BoB - Duel to the Death
image

jeremylarner:
Closes #3171


jeremylarner:
Yes it will.  I think the prompt will also show the card ability being used, but it might be worth checking that that is the case...

jeremylarner:
Needs #3191 to pass tests

PR #3189, opened 2019-05-10
FtE - Purity of spirit

### LINKED ISSUE #3143: FTE - Purity of Spirit

Matthew-Griffith:
closes #3143


server/game/cards/09.1-FtE/PurityOfSpirit.js
```diff
@@ -0,0 +1,29 @@
...
+                cardType: CardTypes.Character,
+                cardCondition: card => card.hasTrait('bushi') && card.isParticipating(),
+                gameAction: AbilityDsl.actions.multiple([
+                    AbilityDsl.actions.honor(),
+                    AbilityDsl.actions.delayedEffect({
+                        when : {
+                            onConflictFinished: () => true
+                        },
+                        gameAction: AbilityDsl.actions.discardStatusToken(context => ({ target: context.target.personalHonor })),
+                        message: 'remove the status token from {0}'
```
jamesoddy: This isn't going to show a very good message in chat.  You probably want something like:
```
message: '{2} removes the {3} from {1} due to the delayed effect of {0}',
messageArgs: context => [context.player, context.target.personalHonor]
```

Matthew-Griffith: so it isn't possible to pass in the `messageArgs` to the function. the interface to `DelayedEffectActionProperties` only defines when, message and gameAction. so I assume this was the intent when the delay effect action was made. I did change the message to 
`'the status token has been removed from {1} due to the delayed effect of {0}'`
let me know if  this is a good message given that the message will only get an array with 'Purity of Spirit' and the name of the character effected

server/game/cards/09.1-FtE/PurityOfSpirit.js
```diff
@@ -0,0 +1,30 @@
...
+                cardCondition: card => card.hasTrait('bushi') && card.isParticipating(),
+                gameAction: AbilityDsl.actions.multiple([
+                    AbilityDsl.actions.honor(),
+                    AbilityDsl.actions.delayedEffect({
+                        when : {
+                            onConflictFinished: () => true
+                        },
+                        gameAction: AbilityDsl.actions.discardStatusToken(context => ({ target: context.target.personalHonor })),
+                        message: '{2} removes the {3} from {1} due to the delayed effect of {0}',
+                        messageArgs: context => [context.player, context.target.personalHonor]
```
jeremylarner: Have you checked to make sure this works in chat?  I don't think we've ever sent a token direct to chat before....

jamesoddy: It's in the tests:
```
expect(this.getChatLogs(5)).toContain('player1 removes the Honored Token from Shiba Tsukune due to the delayed effect of Purity of Spirit');
```
Are we bothered about the capitalisation on "Honored Token"?

jamesoddy:
Re-done on #3219

PR #3190, opened 2019-05-11
The Perfect Gift chat message

Kwaice:
Currently shows that the player reveals its deck twice. Bit misleading.

the perfect gift


jeremylarner:
It's a bug, but I don't think this fixes it.  I think it's a bug in the look at game action...

PR #3193, opened 2019-05-13
Bugfix - Dispatch to no longer target (no choose in ability)

### LINKED ISSUE #2779: Dispatch bug
Dispatch currently requires choosing a target, whereas the card doesn't require a target to be chosen.

jamesoddy:
Closes #2779

### DIFF
--- a/server/game/GameActions/ConditionalAction.ts
+++ b/server/game/GameActions/ConditionalAction.ts
@@ -2,7 +2,7 @@ import { GameAction, GameActionProperties } from './GameAction';
 import AbilityContext = require('../AbilityContext');
 
 export interface ConditionalActionProperties extends GameActionProperties {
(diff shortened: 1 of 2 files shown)


jeremylarner:
Now I look at this again, I see that changing it to be conditional is not as straightforward as I originally thought, because it depends on the target, which isn't accessible outside the game action.  One way to fix this would be to make the game action properties accessible to the condition function - any thoughts?

(so you'd have `condition: (context, properties) => properties.target.inConflict`)

jamesoddy:
Yes, I've been looking too and had exactly the same problem and figured on exactly the same solution.  I've had to use `condition: (context, properties) => properties.target[0].inConflict` which works, but seems a bit hacky to me - what do you think?

jeremylarner:
That is ugly :(
I can't think of a better way of doing it though....

PR #3203, opened 2019-05-22
Bugfix - Illustrious Forge

### LINKED ISSUE #3202: Illustrious Forge and Talisman interaction
image

image

Illustrious forge doesn't seem to be playing nicely with Talisman - needs looking at.

jamesoddy:
Closes #3202

I've added a cardCondition to CardMenuAction so that card buttons can be displayed but disabled (similar to the deck search).

I have also added that CardMenuAction has a legal target if any `handlers` have been added (even if all cards are invalid targets) so that you still receive the prompt.  This means that I needed to adjust Peasant's advice and also a test with Chukan Nobue and Upholding Authority.

Peasant's adivce now uses a CardSelectAction instead of CardMenuAction as it seems a more appropriate fit.  I did have to write a rather ugly cardCondition on Peasant's Advice to prevent it targeting a faceup with province with only facedown dynasty cards in it, but because the discard is optional it was not able to handle it correctly otherwise.

### DIFF
--- a/server/game/cards/05-UotE/PeasantsAdvice.js
+++ b/server/game/cards/05-UotE/PeasantsAdvice.js
@@ -11,7 +11,6 @@ class PeasantsAdvice extends DrawCard {
             target: {
                 cardType: CardTypes.Province,
                 location: Locations.Provinces,
-                cardCondition: card => card.facedown || card.controller.getSourceList(card.location).some(provinceCard => provinceCard.isDynasty && !provinceCard.facedown),
                 gameAction: AbilityDsl.actions.sequential([
                     AbilityDsl.actions.lookAt(),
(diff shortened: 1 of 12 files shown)


server/game/cards/05-UotE/PeasantsAdvice.js
```diff
@@ -11,13 +11,17 @@ class PeasantsAdvice extends DrawCard {
             target: {
                 cardType: CardTypes.Province,
                 location: Locations.Provinces,
+                cardCondition: card => card.facedown || card.controller.getSourceList(card.location).some(provinceCard => provinceCard.isDynasty && !provinceCard.facedown),
```
jeremylarner: Is this necessary? It seems like it shouldn't be....

PR #3216, opened 2019-05-24
Bugfix - War Dog Master

jamesoddy:
`context.costs.discardCard` is returning an array instead of a single target so needs to be `context.costs.discardCard[0]`

### DIFF
--- a/test/server/cards/02.3-ItFC/WarDogMaster.spec.js
+++ b/test/server/cards/02.3-ItFC/WarDogMaster.spec.js
@@ -1,0 +1,43 @@
+describe('War Dog Master', function() {
+    integration(function() {
+        describe('War Dog Master\'s ability', function() {
+            beforeEach(function() {
+                this.setupTest({
+                    phase: 'conflict',
+                    player1: {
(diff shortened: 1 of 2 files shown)


test/server/cards/02.3-ItFC/WarDogMaster.spec.js
```diff
@@ -0,0 +1,45 @@
...
+                this.initiateConflict({
+                    attackers: [this.warDogMaster]
+                });
+                expect(this.player1).toHavePrompt('Triggered Abilities');
+                this.player1.clickCard(this.warDogMaster);
+                expect(this.getChatLogs(2)).toContain('player1 uses War Dog Master, discarding Favorable Ground to give War Dog Master +0military');
+                expect(this.warDogMaster.getMilitarySkill()).toBe(2);
+            });
+        });
+    });
```
jeremylarner: Might be worth having a test that it won't let you resolve it when your D deck is empty

jamesoddy: Good idea.  Done.

PR #3220, opened 2019-05-25
Optional abilities shouldn't get a pass for legal targets

jeremylarner:
Also I added the ability to flag any game action as optional.  Note that there's no logic to back that up however (except for some actions which already support an optional property), so that will need adding in the future.


jamesoddy:
Now included in #3203

PR #3227, opened 2019-05-29
Fix Hida Kisada bug

### LINKED ISSUE #3196: Hida Kisada and Defend Your Honor
When Defend Your Honor (possibly other event cancels but I have only seen Defend Your Honor for this interaction) is used to cancel a Triggered Action from an event's effects, Hida Kisada will still cancel the second Triggered Action's effects. Nothing I can find suggests that cancelling a triggered ability's effects also cancel's a triggered ability's "triggered" state.

Relevant from the Rules Reference page 8 under Event Cards, the bold is my emphasis: "◊ If the effects of an event card are canceled, the card is still considered to have been played, and its costs remain paid, and the card is still discarded. **Only the effects have been canceled.**"

jeremylarner:
I think this fixes #3196, but I haven't tested it

### DIFF
--- a/server/game/cards/01-Core/HidaKisada.js
+++ b/server/game/cards/01-Core/HidaKisada.js
@@ -4,33 +4,34 @@ const { Locations, CardTypes, EventNames, AbilityTypes } = require('../../Consta
 
 class HidaKisada extends DrawCard {
     setupCardAbilities() {
-        this.canCancel = false;
+        this.firstActionEvent = null;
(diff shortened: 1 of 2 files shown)


jamesoddy:
I'll write some tests for Kisada and check this against them later today (probably 😛 )

jamesoddy:
I still need to write those tests 😝 😝 😝 😝

jeremylarner:
Tests added, should be ready to merge

PR #3228, opened 2019-05-29
Bugfix - Feral Ningyo

jamesoddy:
Allow Feral Ningyo to be triggerable from play

Used the nested chat message commit as well to give a better chat message and needed to add `location:` to `DelayedEffectAction` in order to be able to affect Feral Ningyo when in hand.

This could give rise to a strange interaction if there is ever a "cannot put cards in to play into a conflict" or similar effect where Feral Ningyo would be triggerable from hand and not come into play but just setup a delayed effect.  Although currently the returnToDeck action will not return to deck if the card is not in play, so a bit of a strange one to possibly watch out for in the future.

### DIFF
--- a/server/game/GameActions/DelayedEffectAction.ts
+++ b/server/game/GameActions/DelayedEffectAction.ts
@@ -17,23 +17,22 @@ export class DelayedEffectAction extends CardGameAction {
     eventName = EventNames.OnEffectApplied;
     effect = 'apply a delayed effect to {0}';
 
-    defaultProperties: DelayedEffectActionProperties;
+    defaultProperties: DelayedEffectActionProperties = {
+        when: null,
+        message: null,
+        gameAction: null,
+        location: Locations.PlayArea
+    };
+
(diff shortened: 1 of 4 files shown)


server/game/GameActions/DelayedEffectAction.ts
```diff
@@ -21,9 +22,18 @@ export class DelayedEffectAction extends CardGameAction {
...
+        if(properties.location) {
+            let inLocation = Array.isArray(properties.location) ? properties.location.includes(card.location) : properties.location === card.location;
+            if(!inLocation) {
+                return false;
+            }
+        }
+        else {
+            if(card.location !== Locations.PlayArea) {
+                return false;
+            }
```
jeremylarner: There's a number of things about the way you've done this that I don't really like (but it could just be personal preference):
 - For what you're trying to do, I'd rather see you using `defaultProperties` than hard coding default behaviour like this
 - Your location property will get passed to the DelayedEffect too - I'm not sure that an Array is the right thing to pass here, maybe it should be the current location of the card?

PR #3253, opened 2019-06-03
Bugfix - Alibi Artist

jamesoddy:
Fix for Alibi Artist to handle only 1 or 2 cards in deck.

Added a "Take Nothing" choice if there is only one card that will then only return it to the deck.


jeremylarner:
This really needs a game action writing for it, but this will at least fix the bug

PR #3256, opened 2019-06-04
FtE - False loyalities

### LINKED ISSUE #3242: FtE Card #20 : False Loyalties
fte20

Matthew-Griffith:
I can't find the bug for why this card doesn't trigger. I have tried a few things, and if the `afterConflict` condition is changed to just return true and it still doesn't trigger.

Closes #3242 (added by James Oddy)

### DIFF
--- a/server/game/cards/09.1-FtE/FalseLoyalties.js
+++ b/server/game/cards/09.1-FtE/FalseLoyalties.js
@@ -1,8 +1,9 @@
 const DrawCard = require('../../drawcard.js');
 const { Players, CardTypes } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
 
 class FalseLoyalties extends DrawCard {
-    setupCardAbilities(ability) {
+    setupCardAbilities() {
         this.reaction({
             title:'Switch 2 characters your opponent controls',
             when: {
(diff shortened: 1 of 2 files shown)


test/server/cards/09.1-FtE/FalseLoyalities.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+            it('should trigger after a opponent wins a conflict', function() {
+                this.initiateConflict({
+                    type: 'political',
+                    attackers: [this.shono],
+                    defenders: []
+                });
+                this.noMoreActions();
+                expect(this.player2).toHavePrompt('Any Reactions');
+                expect(this.player2).toBeAbleToSelect('false-loyalities');
```
jamesoddy: typos here:
`'Any Reactions'` -> `'Any reactions?'` or `'Triggered Abilities'`
`'false-loyalities'` -> `'false-loyalties'`

test/server/cards/09.1-FtE/FalseLoyalities.spec.js
```diff
@@ -0,0 +1,38 @@
...
+            });
```
jamesoddy: Yeah it isn't triggering because Matsu Berserker has a dash in political so is ineligble to be moved to the conflict so the targeting requirement fails.
BTW, there's a console log you can uncomment in `triggeredability.js` that can sometimes give you a clue as to why something is not triggering.

Matthew-Griffith: thanks I fixed those bugs and added more test and the card seems to be working now

server/game/cards/09.1-FtE/FalseLoyalties.js
```diff
@@ -0,0 +1,42 @@
...
+const AbilityDsl = require('../../abilitydsl');
+
+class FalseLoyalties extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            title:'Switch 2 characters your opponent controls',
+            when: {
+                afterConflict: (event, context) => {
+                    return event.conflict.winner === context.player.opponent &&
+                    context.player.honor < context.player.opponent.honor;
```
jamesoddy: This will give a null reference exception in single player can you please adjust to:

```js
return context.player.opponent && event.conflict.winner === context.player.opponent &&
    context.player.honor < context.player.opponent.honor;
```

PR #3257, opened 2019-06-06
New Feature - Card Stats Popup

jamesoddy:
Adds a new card statistics popup:

image

In order to achieve this, I completely refactored the `getMilitarySkill, getPoliticalSkill, ...` functions on `drawcard.js`.  So people may want to take a look at that to check for any bugs :/ (all tests are passing though).

Also added a new option in profile to disable this functionality if people do not like it.


NoahTheDuke:
Damn, this is tight.

The-Outcast:
Rest looks good! Something in me wants to center the combined values, I though I did that though....

The ordered list display looks real nice for the modifiers. Would have liked some semantic html being applied and it to be an actual ul or ol 😛

jamesoddy:
By all means run some of your "front-end magic" over it.  I'm very unsure of best practices when it comes to front-end so just kind of try things until it works.

And I'm particularly unfamiliar with react so getting the functions to return what I wanted was a bit of a learning experience :)

I'm not precious about it so if you have some changes you want to PR then by all means do. :D

jamesoddy:
The main thing I wanted was to "line-up" the detailed information.  At first I tried a fixed-width font, but it really didn't suit the rest of the aesthetic of the site so resorted to divs, but you may be able to show me a much better way!

PR #3261, opened 2019-06-10
Bugfix - Shinjo Tatsuo to target himself correctly

jamesoddy:
Includes tests


jeremylarner:
Hmmm, this matches the card, but is it the way we want to go in terms of UI?  I'm wondering if a better fix is just to add `condition: context => !context.source.isParticipating()`

jamesoddy:
I'm thinking Hantei - doesn't it need to "target" for him to be able to override?

jamesoddy:
Ah looks like I need to check some other tests as well.  I'll take a look at them.

jeremylarner:
On second thoughts, I think your fix here is the best way to do it

PR #3267, opened 2019-06-12
FtE - Shameless Gossip

### LINKED ISSUE #3236: FtE Card #11 : Shameless Gossip
fte11

jamesoddy:
Closes #3236

I've added a new PersonalHonorStatus enum that makes the implementation of this card a lot simpler o any thoughts on that appreciated.

The best way to get this to work with our current targeting system is to force the first target to be the one with a status token of some sort then the second prompt will force you to choose a character controlled by the same player as the first and without the same honor status.

### DIFF
--- a/server/game/cards/09.1-FtE/ShamelessGossip.js
+++ b/server/game/cards/09.1-FtE/ShamelessGossip.js
@@ -1,6 +1,6 @@
 const DrawCard = require('../../drawcard.js');
 const AbilityDsl = require('../../abilitydsl');
-const { Players, CardTypes, PersonalHonorStatus } = require('../../Constants');
+const { Players, CardTypes } = require('../../Constants');
 
 class ShamelessGossip extends DrawCard {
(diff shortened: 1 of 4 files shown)


server/game/cards/09.1-FtE/ShamelessGossip.js
```diff
@@ -0,0 +1,40 @@
...
+                    cardCondition: card => card.personalHonorStatus !== PersonalHonorStatus.Ordinary
+                },
+                second: {
+                    activePromptTitle: 'Choose a Character to move the status token to',
+                    dependsOn: 'first',
+                    cardType: CardTypes.Character,
+                    cardCondition: (card, context) =>
+                        card.controller === context.targets.first.controller &&
+                        card !== context.targets.first &&
+                        card.personalHonorStatus !== context.targets.first.personalHonorStatus,
```
jeremylarner: This line shouldn't be needed

PR #3272, opened 2019-06-12
Bugfix - Add null checks

jamesoddy:
Some checks to prevent some null reference exceptions

### DIFF
--- a/server/game/AbilityTargets/AbilityTargetSelect.js
+++ b/server/game/AbilityTargets/AbilityTargetSelect.js
@@ -43,7 +43,7 @@ class AbilityTargetSelect {
     }
 
     getGameAction(context) {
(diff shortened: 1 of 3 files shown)


server/game/AbilityTargets/AbilityTargetSelect.js
```diff
@@ -43,7 +43,7 @@ class AbilityTargetSelect {
     }
 
     getGameAction(context) {
-        let choice = this.properties.choices[context.selects[this.name].choice];
+        let choice = this.properties.choices[context.selects[this.name] && context.selects[this.name].choice];
```
jeremylarner: if there's no select, don't you want to return an empty array here?

jamesoddy: Sure.  I've adjusted as requested.

PR #3273, opened 2019-06-12
FtE - Gateway to Meido

### LINKED ISSUE #3233: FtE Card #3 : Gateway to Meido
fte3

jamesoddy:
Closes #3233


jamesoddy:
I've got this to work with disguised but can't say I'm 100% happy with it.  It's a bit of a messy implementation and not flexible (if say we needed a play only at home effect in the future).

PR #3276, opened 2019-06-17
FtE - Give No Ground

### LINKED ISSUE #3237: FtE Card #14 : Give No Ground
fte14

jamesoddy:
Closes #3237

No ruling, but this is best guess at the ultimate ruling (existing lasting effects with negative modifiers will not affect skills).  I will adjust if we hear different :)


jamesoddy:
Thinking about it, this needs more thought.  I'm not sure if new lasting effects are allowed to be triggered as per the latest "set" ruling.

jamesoddy:
Changed to not retrospectively affect existing lasting effects.

I pulled in Festival of the Departed implementation in order to use the `CannotApplyLastingEffects` effect.

jamesoddy:
In the middle of re-working for new rulings (including Festival of the Departed) so don't merge/review at the moment.

server/game/cards/09.1-FtE/GiveNoGround.js
```diff
@@ -0,0 +1,32 @@
...
+
+class GiveNoGround extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Increase a character\'s military skill',
+            condition: () => this.game.isDuringConflict(ConflictTypes.Military),
+            target: {
+                cardType: CardTypes.Character,
+                cardCondition: card => card.isDefending(),
+                gameAction: AbilityDsl.actions.multiple([
```
jeremylarner: do you need multiple here? Can't you just pass an array of effects?

server/game/GameObject.js
```diff
@@ -23,8 +23,18 @@ class GameObject {
...
 
     getEffects(type) {
-        let filteredEffects = this.effects.filter(effect => effect.type === type);
-        return filteredEffects.map(effect => effect.getValue(this));
+        let suppressedEffects = this.getSuppressedEffects();
+        if(type) {
+            let filteredEffects = this.effects.filter(effect => effect.type === type && !suppressedEffects.includes(effect));
+            return filteredEffects.map(effect => effect.getValue(this));
+        }
+        return this.effects.filter(effect => !suppressedEffects.includes(effect));
```
jeremylarner: I don't really like that you're returning an array of raw effects when there's no type specified....
I don't know whether we ever should call getEffects with no type, but I think should be at least mapping them to their values

server/game/GameObject.js
```diff
@@ -23,8 +23,18 @@ class GameObject {
...
+    }
+
+    getSuppressedEffects() {
+        let suppressingEffects = this.effects.filter(effect => effect.type === EffectNames.SuppressEffects);
+        let suppressedEffects = this.effects.filter(effect => suppressingEffects.some(suppressingEffect => suppressingEffect.value.value(effect)));
+        return suppressedEffects;
     }
 
     sumEffects(type) {
```
jeremylarner: sumEffects and anyEffect need updating to use the new getEffects

server/game/basecard.ts
```diff
@@ -66,7 +66,7 @@ class BaseCard extends EffectSource {
     get actions(): CardAction[] {
         let actions = this.abilities.actions;
         if(this.anyEffect(EffectNames.CopyCharacter)) {
```
jeremylarner: Ah, now your above code makes sense.  Yeah, I think we need to fix this hack, not hack something else to make this work....

server/game/StatModifier.ts
```diff
@@ -0,0 +1,65 @@
+import { CardTypes } from "./Constants";
+
+class StatModifier {
```
jeremylarner: I'm not really sure what this is for - is it just to make your stats display work nicely?

server/game/drawcard.js
```diff
@@ -512,6 +507,7 @@ class DrawCard extends BaseCard {
     }
 
     setPersonalHonor(token) {
+        !token && this.personalHonor && this.personalHonor.setCard(null);
```
jeremylarner: I'm really not a fan of this syntax....

server/game/StatusToken.js
```diff
@@ -1,12 +1,47 @@
...
         this.honored = !!isHonored;
         this.dishonored = !isHonored;
         this.card = card;
         this.type = 'token';
+        this.persistentEffects = [];
+
+        this.applyHonorEffects();
+    }
+
+    applyHonorEffects() {
```
jeremylarner: My instinct is that this is in the wrong place, and that it should be dealt with in either basecard or drawcard with the rest of the effect housekeeping.  It looks like you're messing with effects outside of `checkGameState` and that's going to mess up how state-based effects work.....

server/game/Effects/Effect.js
```diff
@@ -88,6 +88,9 @@ class Effect {
         if(this.duration !== Durations.Persistent) {
             return true;
         }
+        if(this.source.type === 'token') {
+            return true;
+        }
         let effectOnSource = this.source.persistentEffects.some(effect => effect.ref && effect.ref.includes(this));
```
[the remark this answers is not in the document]
jamesoddy: Sure, this was a bit of hack to get it to work, but yeah I suppose tokens are never 'facedown'.  Nice idea about Yojiro, that may save having the HonorStatusDoesNotModifySkill effect.  I suggest that is saved for another PR though.

server/game/Effects/Effect.js
```diff
@@ -127,7 +130,7 @@ class Effect {
         return {
             source: this.source.name,
             targets: _.map(this.targets, target => target.name),
-            active: this.duration !== Durations.Persistent || !this.source.isBlank(),
```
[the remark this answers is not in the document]
jamesoddy: This is just the getDebugInfo() function so it just seemed to make sense to use the isEffectActive function.  Not a big deal either way.

server/game/GameObject.js
jamesoddy: In that case I would prefer that we call this function getEffectValues() and have a separate function for getEffects().  With suppressed effects I need a function that returns all of the effects minus any suppressed effects.  Unfortunately this function was getting in the way of that as it was named getEffects().

server/game/GameObject.js
jamesoddy: 👍

server/game/StatModifier.ts
jamesoddy: Yeah this is just to make the generation of the modifiers a bit simpler in `drawcard.js`

server/game/StatusToken.js
jamesoddy: I'm sure you are right, I will see what I can do.  I originally wanted the "source" of the effect to be the token, I suppose I can still do that via the drawcard though :)

server/game/basecard.ts
jamesoddy: I'll refactor it all and see if I can simplify it.

server/game/drawcard.js
```diff
@@ -93,6 +96,26 @@ class DrawCard extends BaseCard {
...
+        let politicalBonus = parseInt(this.cardData.political_bonus);
+        if(politicalBonus) {
+            this.persistentEffect({
+                condition: () => true,
+                match: (card) => card === this.parent,
+                targetController: Players.Any,
+                effect: AbilityDsl.effects.modifyPoliticalSkill(politicalBonus)
+            });
+        }
+    }
```
jamesoddy: @jeremylarner Did you see this part, does this look reasonable for creating attachment bonuses as effects?

server/game/Effects/Effect.js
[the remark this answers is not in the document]
jamesoddy: Ah yeah, I put this in because the source (the token) does not have any persistent effects.  I think I will try and reverse all of this and have a go at adding the persistent effects directly on to the card.

server/game/GameObject.js
jamesoddy: Actually I'll add a new function `getRawEffects()` to return the unsupressed raw effects.

server/game/Effects/Effect.js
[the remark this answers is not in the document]
jeremylarner: Yeah, it's not your change I was worried about, more that I couldn't find anything checking for blank effects....

server/game/GameObject.js
jeremylarner: That's fine - I think generally it would be better to avoid getRawEffects() as much as possible.  It will make future refactors of the effects system much easier...

server/game/drawcard.js
jeremylarner: Yes, this looks fine

jeremylarner:
In order to complicate this even more, we need to make sure it correctly interacts with base skill modifiers (like WotL) and switches (like Bayushi Yunako).  rules_advanced seems to think that WotL should be suppressed by GnG, but switches should not

server/game/Effects/StaticEffect.js
```diff
@@ -45,7 +59,7 @@ const conflictingEffects = {
 };
 
 class StaticEffect {
-    constructor(type = '', value) {
```
jeremylarner: why did you change this?

server/game/StatusToken.js
```diff
@@ -1,12 +1,46 @@
...
+            return;
+        }
+        let amount = this.honored ? card => card.getGlory() : card => 0 - card.getGlory();
+        let effect = {
+            match: this.card,
+            effect: AbilityDsl.effects.modifyBothSkills(amount)
+        };
+        this.persistentEffects.push(effect);
+        effect.ref = this.addEffectToEngine(effect);
+        this.game.effectEngine.checkEffects();
```
jeremylarner: I played with this for a while before working out that the way you've done it is indeed correct.  However, I don't think this `checkEffects()` should be here, otherwise you risk updating state-based effects in the wrong place.  Removing it will make a couple of tests fail, but that's because those tests are lazy (they are just calling 'card.dishonor()` instead of finding a real effect to make that game state change).

server/game/StatusToken.js
```diff
@@ -1,12 +1,46 @@
...
+    }
+
+    applyHonorEffects() {
+        if(!this.card) {
+            return;
+        }
+        let amount = this.honored ? card => card.getGlory() : card => 0 - card.getGlory();
+        let effect = {
+            match: this.card,
+            effect: AbilityDsl.effects.modifyBothSkills(amount)
```
jeremylarner: This needs to be a dynamic effect, otherwise changes in glory won't be reflected

server/game/Effects/StaticEffect.js
jamesoddy: Actually I think that was you as part of the Festival of the Departed PR 😛

server/game/StatusToken.js
jamesoddy: Yeah that was basically it.  I'll try removing it and see if I can fix the tests.

jeremylarner: The easy fix for the tests is to just call `checkEffects()` after `card.dishonor()`

server/game/StatusToken.js
jamesoddy: modifyBothSkills is a `flexible` effect and I am passing a function so it should be dynamic - unless I am missing what you are saying.

jeremylarner: yes, I just can't read.... :)

PR #3281, opened 2019-06-21
Byoyomi - fixes and additional options

jamesoddy:
Fixes the byoyomi clock to work.

Now resets the clock when either:
- the current gamestep completes (resets both player's clocks) or 
- the player is presented with a waiting prompt (resets the waiting player's clock)

Now allows you to specify a main timer and a number of byoyomi periods and time period (10 mins + 3 x 30 secs for example).


jamesoddy:
image
image
image

Issue #3238, opened 2019-05-31
FtE Card #15 : Moment of Perfect Beauty


gryffon:
![fte15](https://user-images.githubusercontent.com/6053185/58711167-e52af000-838b-11e9-919d-731267ae9386.png)

Matthew-Griffith:
I'll take this one

jeremylarner:
This one is also hard. I would do it by adding a player effect, and then checking for it in the Action window

ngroover:
I can try taking this one.

jamesoddy:
Waiting on ruling regarding this vs. Hawk Tattoo

PR #3282, opened 2019-06-24
FtE - Moment of Perfect Beauty (#3238)

ngroover:


test/server/cards/09.1-FtE/MomentOfPerfectBeauty.spec.js
```diff
@@ -0,0 +1,94 @@
...
+                this.initiateConflict({
+                    attackers: ['doji-challenger'],
+                    defenders: ['adept-of-the-waves'],
+                    type: 'political'
+                });
+                this.dojichallenger.honor();
+                this.player2.pass();
+                this.player1.clickCard(this.momentofbeauty);
+                this.player2.clickCard(this.finekatana);
+                expect(this.player1).not.toHavePrompt('Conflict Action Window');
```
jeremylarner: This test doesn't work - player2 is currently being prompted to choose a character to play Katana on.  As a general rule, check for the prompt you expect the player to have.

server/game/gamesteps/actionwindow.js
```diff
@@ -116,7 +116,18 @@ class ActionWindow extends UiPrompt {
     pass() {
         this.game.addMessage('{0} passes', this.currentPlayer);
 
-        if(this.prevPlayerPassed || !this.currentPlayer.opponent) {
+        let resolveConflictEarlyEffects = this.currentPlayer.getEffects(EffectNames.ResolveConflictEarly);
+        var resolveConflictEarly = false;
+        for(let i = 0; i < resolveConflictEarlyEffects.length; i++) {
```
jeremylarner: This is only going to countdown when a player passes.  I'm not sure the idea of decrementing a counter will work generally in this case, because I don't think you can change the value stored in the effect in this way.

ngroover: I was trying to allow specifying a number of turns left to make the effect a little more generic in case it's needed later. Would you recommend just checking for the effect itself and ending the conflict? Or should I move tracking the turns left to part of the player itself so I can modify it?

jeremylarner: I think if you want to do a counter, you probably need to track it in the action window itself. To be honest though, I think it's probably a case of overengineering, and you should just check for the effect and end the conflict.  If we get something which needs a tracker, we'll come up with a solution at that point.

server/game/cards/09.1-FtE/MomentOfPerfectBeauty.js
```diff
@@ -0,0 +1,23 @@
...
+            title: 'One more action and then end the conflict',
+            condition: context => this.game.isDuringConflict() &&
+                    this.game.currentConflict.getNumberOfParticipantsFor(context.player, card => card.isHonored) >
+                    this.game.currentConflict.getNumberOfParticipantsFor(context.player.opponent, card => card.isHonored),
+            gameAction: AbilityDsl.actions.playerLastingEffect({
+                duration: Durations.UntilEndOfConflict,
+                targetController: Players.Opponent,
+                effect: AbilityDsl.effects.resolveConflictEarly()
+            })
+        });
```
jamesoddy: You could do with an `effect:` here otherwise you will not get a very meaningful chat message.  Something like:

`effect: 'resolve the conflict after {1]'s next action',
`effectArgs`: context => [context.player.opponent]

PR #3284, opened 2019-06-28
FtE - Sententious Poet

### LINKED ISSUE #3214: FTE - Sententious Poet
image

jamesoddy:
Closes #3214 

With no reponse for a ruling to the timing, I have set it to be after step 8 as it was the easiest to implement 😛

### DIFF
--- a/server/game/gamesteps/abilityresolver.js
+++ b/server/game/gamesteps/abilityresolver.js
@@ -115,6 +115,7 @@ class AbilityResolver extends BaseStepWithPipeline {
             this.cancelled = true;
             return;
         }
+        this.preCostsFate = this.context.player.fate;
         if(this.costEvents.length > 0) {
             this.game.openEventWindow(this.costEvents);
         }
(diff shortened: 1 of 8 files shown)


server/game/gamesteps/abilityresolver.js
```diff
@@ -128,6 +129,8 @@ class AbilityResolver extends BaseStepWithPipeline {
         if(this.cancelled) {
             this.game.addMessage('{0} attempted to use {1}, but did not successfully pay the required costs', this.context.player, this.context.source);
         }
+        let chooseFate = this.context.chooseFate === undefined ? 0 : this.context.chooseFate;
+        this.context.spentFate = this.preCostsFate - this.context.player.fate - chooseFate;
```
jeremylarner: This isn't a great way to do this long term.  You're going to catch any fate spent by reactions to cost events, which shouldn't trigger the Poet

jamesoddy: I was trying to pick up any additional costs that may come up (such as the upcoming Magistrate of Reason).  Previously I was looking at just `event.context.costs.fate > 0 || event.context.costs.targetDependentFate > 0` but figured that wouldn't pick up any additional costs.  I suppose this will just have to adjust once I know how Magistrate of Reason will work.

jeremylarner: The correct way to do it is to change fate costs to generate events properly, and then `when` of those events

jeremylarner: For righteous magistrate, I guess it's already using an event correctly, so it should `when` from that... I guess I should also say that I suspect the ruling will be that it triggers in step 4, so I'm thinking of solutions for that.  I don't know if there's any way to access what costs were paid in step 8.  If that's the way we want to go, we should probably create a costEvent array on the context object

PR #3285, opened 2019-06-28
FtE - Greater Understanding

The-Outcast:
Still have some test left to write, but code-wise this is my first submission.


.vscode/launch.json
```diff
@@ -1,6 +1,14 @@
...
     "version": "0.2.0",
     "configurations": [
+        {
+            "type": "node",
+            "request": "attach",
+            "name": "Attach by Process ID",
+            "processId": "${command:PickProcess}",
+            "restart": true,
+            "protocol": "inspector",
+        },
```
NoahTheDuke: Is this intentionally included?

The-Outcast: No that seems like my own debug configuration that was added. Ill fix it

PR #3319, opened 2019-07-08
Bugfix - Zentaro

jamesoddy:


NoahTheDuke:
What's the reasoning for removing `/cancel-prompt`?

jamesoddy:
"/cancel-prompt is a legacy command that isn't really supported any more"
"manual mode to cancel the prompt should cancel the prompt in a much more elegant way"

Essentially, people should not be using cancel-prompt so it seems best to take it out of the system as it may be causing a number of crashes on the server.

NoahTheDuke:
Cool, that makes total sense. Thanks!

server/game/PlayDisguisedCharacterAction.js
```diff
@@ -48,6 +49,9 @@ class DisguisedReduceableFateCost extends ReduceableFateCost {
     }
 
     getReducedCost(context) {
+        if(!context.costs.chooseDisguisedCharacter) {
+            return;
```
jeremylarner: I'm not sure what this is intended to achieve - you're just pushing a potential crash somewhere else...

PR #3343, opened 2019-07-12
BoB - Fury of the Damned

### LINKED ISSUE #3328: BoB - Fury of the Damned #33
Fire role only. Action: During a conflict, choose any number of participating Bushi characters you control - double the base MIL skill of each of those characters until the end of the conflict. At the end of the conflict, sacrifice each of those characters.

jamesoddy:
Closes #3328 

Also changed to `ModifyBaseMilitarySkillMultiplier` and `ModifyBasePoliticalSkillMultiplier`


jeremylarner:
This looks fine - my only concern is that you've coded the delayed effect as a single effect which sacrifices all characters simultaneously, rather than multiple delayed effects which sacrifice each character one at a time.  I don't know which is correct (I think we'd need to ask Tyler)

server/game/cards/09.2-BoB/FuryOfTheDamned.js
```diff
@@ -0,0 +1,40 @@
...
+const AbilityDsl = require('../../abilitydsl');
+
+class FuryOfTheDamned extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Double the base military skill',
+            condition: context => context.game.isDuringConflict(),
+            target: {
+                activePromptTitle: 'Choose bushi characters',
+                numCards: 0,
```
jeremylarner: I would prefer to see `mode: TargetModes. Unlimited` here rather than `numCards: 0`
It's more forward compatible, and easier to understand

jamesoddy: Done

PR #3347, opened 2019-07-15
BoB - Benevolent Ambassador #31

### LINKED ISSUE #3327: BoB - Benevolent Ambassador #31
Courtesy (When this character leaves play, gain 1 fate.) Reaction: After this character wins a conflict - you and your opponent each gain 1 honor.

Matthew-Griffith:
closes #3327


test/server/cards/09.2-BoB/BenevolentAmbassador.spec.js
```diff
@@ -0,0 +1,66 @@
...
+            });
+
+            it('should not trigger if you lose the conflict', function() {
+                this.initiateConflict({
+                    attackers: [this.ambassador],
+                    defenders: [this.maiden]
+                }),
+                this.player2.playAttachment('fine-katana', this.maiden);
+                this.noMoreActions();
+                expect(this.player1).not.toBeAbleToSelect(this.ambassador);
```
jamesoddy: This isn't necessarily a wrong way to test, but it is much more preferable to check if you are actually at the step you expect to be rather than using a `.not.`.  So in this case something like `expect(this.player1).toHavePrompt('Air Ring')` woudl be preferable.

test/server/cards/09.2-BoB/BenevolentAmbassador.spec.js
```diff
@@ -0,0 +1,66 @@
...
+            });
+
+            it('should not trigger if it is at home', function() {
+                this.initiateConflict({
+                    attackers: ['isawa-tadaka'],
+                    defenders: []
+                });
+                this.noMoreActions();
+                expect(this.player1).not.toBeAbleToSelect(this.ambassador);
```
jamesoddy: As above.

PR #3349, opened 2019-07-17
BoB - Hiruma Yoshino

### LINKED ISSUE #3322: BoB - Hiruma Yoshino #26
Disguised Bushi. Action: During a MIL conflict in which this character is participating, choose a character card in the attacked province - that character contributes its printed MIL skill towards your side during the resolution of this conflict.

jamesoddy:
Closes #3322 

This ended up a bit of a beast of a PR.

All of the "cannot contibute" magistrates have been changed to now affect the conflict and not the cards using the new `cannotContribute` effect.

Any lasting effects on cards are now removed when moving between any locations except between provinces.

Also added a new `changeSkillFunction` effect for cards that will override any requests for getSkill().

Additional contributors to the conflict are now not based upon the controller of the card, but by the player specified as part of the effect.

### DIFF
--- a/server/game/conflict.js
+++ b/server/game/conflict.js
@@ -269,8 +269,13 @@ class Conflict extends GameObject {
 
     calculateSkillFor(cards) {
         let skillFunction = this.mostRecentEffect(EffectNames.ChangeConflictSkillFunction) || (card => card.getSkill(this.conflictType));
+        let cannotContributeFunctions = this.getEffects(EffectNames.CannotContribute);
         return cards.reduce((sum, card) => {
-            if(card.bowed || !card.allowGameAction('countForResolution')) {
+            let cannotContribute = card.bowed;
+            if(!cannotContribute) {
+                cannotContribute = cannotContributeFunctions.some(func => func(card));
+            }
+            if(cannotContribute) {
                 return sum;
             }
             return sum + skillFunction(card);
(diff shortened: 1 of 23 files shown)


server/game/drawcard.js
```diff
@@ -207,13 +207,16 @@ class DrawCard extends BaseCard {
...
+     * @return {number} The chosen skill value
+     */
     getSkill(type) {
-        /**
-         * Direct the skill query to the correct sub function.
-         * @param  {string} type - The type of the skill; military or political
-         * @return {integer} The chosen skill value
-         */
-        if(type === 'military') {
+        let skillFunction = this.mostRecentEffect(EffectNames.ChangeSkillFunction);
```
jeremylarner: I'm not sure that this is the best way to do this.  Shouldn't the engine be looking for ChangeSkillFunction in Conflict?

server/game/conflict.js
```diff
@@ -228,12 +228,25 @@ class Conflict extends GameObject {
             return stateChanged;
         }
 
-        let additionalCharacters = this.getEffects(EffectNames.ContributeToConflict);
+        const contributingLocations = [
+            Locations.PlayArea,
+            Locations.ProvinceOne,
+            Locations.ProvinceTwo,
+            Locations.ProvinceThree,
+            Locations.ProvinceFour
```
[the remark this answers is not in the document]
jamesoddy: Done

server/game/drawcard.js
jamesoddy: I have wrapped the getSkill function with a getContributionToConflict function which will take this effect into consideration so the skill of the cards should not be muddied by the contribution to the conflict.  

(and also changed the name of the effect to `ChangeContributionFunction`)

PR #3365, opened 2019-07-28
Bugfix - Shosuro Sadako and From the Shadows interaction

### LINKED ISSUE #3363: From the Shadows + Sadako bugged
After the release of Bonds of Blood, looks like there's a regression on From the Shadows + Shosuro Sadako. See screenshot.

From the Shadows

jamesoddy:
Closes #3363

~~Not an ideal solution as it now moves the dishonoring to after the card has been moved into play, so any reactions to coming into play will not treat Sadako as dishonored.~~

~~Needs more thought.~~


jamesoddy:
Ok so I have adjusted to reapply the honor/dishonor effect when the card enters play as the lasting effect was being removed when the card moved from out of play into play.

PR #3410, opened 2019-08-12
JfS - Shosuro Ibuki

### LINKED ISSUE #3384: JfS - Shosuro Ibuki
image

The-Outcast:
Closes #3384

Implementation for Shosuro Ibuki

### DIFF
--- a/server/game/cards/09.3-JfS/ShosuroIbuki.js
+++ b/server/game/cards/09.3-JfS/ShosuroIbuki.js
@@ -11,7 +11,7 @@ class ShosuroIbuki extends DrawCard {
                     context.source.isParticipating()
             },
             gameAction: AbilityDsl.actions.removeFate(context => ({
-                target: context.game.currentConflict.getParticipants(participant => participant.uuid != context.source.uuid),
+                target: context.game.currentConflict.getParticipants(participant => participant.uuid !== context.source.uuid),
                 amount: 1
             })),
             effect: 'remove one fate from each other participating character'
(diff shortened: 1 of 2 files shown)


server/game/cards/09.3-JfS/ShosuroIbuki.js
```diff
@@ -0,0 +1,24 @@
...
+            title: 'Remove one fate from each other participating character',
+            when: {
+                afterConflict: (event, context) =>
+                    event.conflict.winner === context.player &&
+                    context.source.isParticipating()
+            },
+            gameAction: AbilityDsl.actions.removeFate(context => ({
+                target: context.game.currentConflict.getParticipants(participant => participant.uuid !== context.source.uuid),
+                amount: 1
+            })),
```
jamesoddy: I'm not 100% sure, but I expect the `.uuid` and `amount: 1` are unnecessary

server/game/cards/09.3-JfS/ShosuroIbuki.js
```diff
@@ -0,0 +1,24 @@
...
+            effect: 'remove one fate from each other participating character'
```
jamesoddy: Hmm, this may be better as 'remove 1 fate from Target1, Target2, ...'
i.e.
`effect: 'remove 1 fate from {0}'`
where I think {0} will be `context.targets` (I'm never quite sure if it's {0} or {1} in this scenario)

jamesoddy: As discussed on discord (some time ago!), we agreed that actually your message is better

PR #3416, opened 2019-08-18
Merge develop into master

jamesoddy:
Includes bugfixes and glory stat popup


server/game/GameActions/AttachAction.ts
```diff
@@ -28,6 +35,8 @@ export class AttachAction extends CardGameAction {
             return false;
         } else if(!properties.attachment || properties.attachment.anotherUniqueInPlay(context.player) || !properties.attachment.canAttach(card, context)) {
             return false;
+        } else if(properties.takeControl && properties.attachment.controller === context.player) {
```
jeremylarner: I'm not sure this is a great idea. An effect which moves an attachment and takes control of it is still a game state change if it cannot take control (e.g. Karada District)

test/server/cards/02.1-ToA/KaradaDistrict.spec.js
```diff
@@ -124,6 +125,8 @@ describe('Karada District', function() {
 
                     expect(this.watchCommander.controller).toBe(this.player2.player);
                     expect(this.miyaMystic.attachments.toArray()).toContain(this.watchCommander);
+                    expect(this.getChatLogs(2)).toContain('player2 uses Karada District, giving 1 fate to player1 to choose a target for Watch Commander');
+                    expect(this.getChatLogs(1)).toContain('player2 chooses to attach Watch Commander to Miya Mystic');
                 });
             });
```
jeremylarner: If this is passing tests, we presumably don't have a test checking that you can move an attachment you control, would it be worth adding one?

server/game/GameActions/AttachAction.ts
```diff
@@ -28,6 +35,8 @@ export class AttachAction extends CardGameAction {
             return false;
         } else if(!properties.attachment || properties.attachment.anotherUniqueInPlay(context.player) || !properties.attachment.canAttach(card, context)) {
```
jeremylarner: I think `anotherUniqueInPlay` only needs checking if we're taking control?

server/game/GameActions/AttachAction.ts
```diff
@@ -28,6 +35,8 @@ export class AttachAction extends CardGameAction {
...
             return false;
+        } else if(properties.takeControl && properties.attachment.controller === context.player) {
+            return false;
         }
         return card.allowAttachment(properties.attachment) && super.canAffect(card, context);
```
jeremylarner: Do we need a check that either the card is changing control or it's changing parent?

server/game/playattachmentaction.js
```diff
@@ -40,7 +40,8 @@ class PlayAttachmentAction extends BaseAction {
             originalLocation: context.source.location,
             playType: PlayTypes.PlayFromHand
         });
-        context.game.openEventWindow([context.game.actions.attach({ attachment: context.source }).getEvent(context.target, context), cardPlayedEvent]);
+        let takeControl = context.source.controller !== context.player;
```
jeremylarner: This looks wrong to me.  I think `takeControl` should always be `true`.  If I play an attachment using Infiltrator, I should control it

jamesoddy: The problem here is that if I set this to `true` in all cases then it interferes with Karada/CiF as they are explicitly checking that you can only "take control" if you are not the current controller.  This will need a bit of an overhaul I think but for now the above still works with Infiltrator as the controller of the card is not the player anyway.  I'll push a test for that though as it is one to watch out for in the future.

server/game/GameActions/AttachAction.ts
jamesoddy: Tests for Gunbai fail unless we do this as it needs to check if another unique is in play but there is no "taking control"

server/game/GameActions/AttachAction.ts
jamesoddy: As discussed, there are currently no effects that independently take control and/or move an attachment.  So we need this still to check for Karada/CiF.

test/server/cards/02.1-ToA/KaradaDistrict.spec.js
jamesoddy: We do have a test, but as discussed, but Karada cannot move an attachment you control (it is just discarded).

PR #3417, opened 2019-08-19
Test - Infiltrator

jamesoddy:
Itty bitty change to see if Infiltrator changes control of attachments when played


jamesoddy:
I'll just merge this in as it's so minor and best to go in with the latest PR to master

PR #3420, opened 2019-08-25
JfS - Conflict Between Kin

### LINKED ISSUE #3372: JfS - Conflict Between Kin
image

jeremylarner:
Untested
Closes #3372

### DIFF
--- a/server/game/cards/09.3-JfS/ConflictBetweenKin.js
+++ b/server/game/cards/09.3-JfS/ConflictBetweenKin.js
@@ -1,0 +1,27 @@
+const ProvinceCard = require('../../provincecard.js');
+const { Players } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl');
+
+class ConflictBetweenKin extends ProvinceCard {
+    setupCardAbilities() {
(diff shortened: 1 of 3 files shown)


server/game/cards/09.3-JfS/ConflictBetweenKin.js
```diff
@@ -0,0 +1,27 @@
...
+class ConflictBetweenKin extends ProvinceCard {
+    setupCardAbilities() {
+        this.persistentEffect({
+            condition: context => context.source.isConflictProvince(),
+            targetController: Players.Opponent,
+            match: card => card.isParticipating(),
+            effect: [
+                AbilityDsl.effects.cardCannot({
+                    cannot: 'target',
+                    restricts: 'events'
```
The-Outcast: This up for interpretation, but I'm not sure 'cannot be chosen as the target of an event...' and 'and cannot have an attachment played on it if that card shares a clan affiliation with the with the character' are two seperate restrictions.

I think both 'events' and 'attachments' can't share a clan affiliation. But other events should still be able to be played.

PR #3454, opened 2019-09-07
CoH - Togashi Yoshi

### LINKED ISSUE #3437: CoH Card - Togashi Yoshi

The-Outcast:
closes #3437

### DIFF
--- a/server/game/cards/09.4-TCoH/TogashiYoshi.js
+++ b/server/game/cards/09.4-TCoH/TogashiYoshi.js
@@ -1,4 +1,3 @@
-import { TargetModes } from '../../Constants.js';
 const DrawCard = require('../../drawcard.js');
 const AbilityDsl = require('../../abilitydsl');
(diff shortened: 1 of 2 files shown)


server/game/cards/09.4-TCoH/TogashiYoshi.js
```diff
@@ -0,0 +1,31 @@
...
+            when: {
+                afterConflict: (event, context) => event.conflict.winner === context.player &&
+                    context.source.isParticipating()
+            },
+            target: {
+                mode: TargetModes.Ring,
+                ringCondition:  ring => ring.fate >= 1 &&
+                    ring.contested === false &&
+                    ring.claimed === false
+            },
```
jamesoddy: This is targeting when there is no choose in the action, this could cause some issues with Hantei 38.  I think we should be using a SelectRingAction.

PR #3474, opened 2019-09-10
CoH - Ide ryoma

### LINKED ISSUE #3444: CoH - Ide Ryoma
image

Matthew-Griffith:
closes #3444


Matthew-Griffith:
@The-Outcast yeah I had a thought last night that I could make the card without a menu prompt but it didn't work. I haven't fully convert the card to using a menu prompt because it was getting late last night so I just pushed it in an unfinished state just in case since I wouldn't be able to work on it for a bit. for the naming  I have never been great at that and would like suggestions.

PR #3487, opened 2019-09-18
CoH - Iron Crane Legion

### LINKED ISSUE #3436: CoH - Iron Crane Legion
image

jamesoddy:
Closes #3436

Added a new effect calculatePrintedMilitarySkill and adjusted the calculation of printed military skill to suit.

### DIFF
--- a/test/server/cards/09.4-TCoH/IronCraneLegion.spec.js
+++ b/test/server/cards/09.4-TCoH/IronCraneLegion.spec.js
@@ -29,7 +29,7 @@ describe('Iron Crane Legion', function() {
                 });
 
                 it('should give Iron Crane Legion a base military skill equal to the number of cards your opponent has in hand', function() {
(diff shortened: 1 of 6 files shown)


test/server/cards/09.4-TCoH/IronCraneLegion.spec.js
```diff
@@ -0,0 +1,44 @@
...
+                beforeEach(function() {
+                    this.noMoreActions();
+                    this.initiateConflict({
+                        attackers: [this.ironCraneLegion],
+                        defenders: []
+                    });
+                });
+
+                it('should give Iron Crane Legion a base military skill equal to the number of cards your opponent has in hand', function() {
+                    expect(this.ironCraneLegion.getBaseMilitarySkill()).toBe(3);
```
The-Outcast: While looking at the code and implementation it is probably correct, but can you write this as opponent.hand.length? So it's more clear it's the opponents hand size that counts and not some random int.

PR #3538, opened 2019-10-06
TEL - Ardent Omoidasu

### LINKED ISSUE #3517: TEL - Ardent Omoidasu
image

The-Outcast:
- Implement Ardent Omoidasu
- Tests for Ardent Omoidasu

closes #3517


server/game/cards/10-TEL/ArdentOmoidasu.js
```diff
@@ -0,0 +1,21 @@
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl');
+
+class ArdentOmoidasu extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            title: 'Steal 2 honor',
+            when: {
+                onCardDishonored: (event, context) => event.card.controller === context.player
+                    && event.context.player === context.player.opponent
```
jeremylarner: I think you should check specifically what the card says (i.e. is this an opponent's card or ring effect?) rather than assuming that checking context.player will cover all eventualities.

tvdstaaij:
Here's an interesting question that I ran into when implementing Akodo Motivator and also applies to Ardent Omoidasu: what happens if a player has Omoidasu and Toturi, and triggers Toturi on defense in a fire conflict? The underlying question here would be: who is the "owner" of a ring effect? I would say that would be the player resolving the ring (regardless of who gets the benefit). So if my interpretation is correct, then Ardent Omoidasu would not be able to trigger in this situation since it's not an opponents' ring effect (compare Ready for Battle which says "a ring effect"). Expressed as a test:

```
        it('should not allow to trigger when dishonored by a fire ring that Toturi resolves on defense', function() {
            this.noMoreActions();
            this.player1.passConflict();
            this.noMoreActions();
            this.initiateConflict({
                type: 'military',
                attackers: [this.mirumotoRaitsugu],
                defenders: [this.akodoToturi],
                ring: 'fire'
            });
            this.noMoreActions();

            expect(this.player1).toBeAbleToSelect(this.akodoToturi);
            this.player1.clickCard(this.akodoToturi);

            this.player2.clickCard(this.ardentOmoidasu);
            this.player2.clickPrompt('Dishonor Ardent Omoidasu');

            expect(this.player1).toHavePrompt('Action Window');
        });
```

Currently, this test fails and Omoidasu is allowed to trigger. This raises two questions:
* Should this test pass or is my interpretation wrong?
* Does the engine allow us to differentiate between the player resolving a ring effect, and the player who is considered the attacking player while resolving the effect, from the context of an event? It seems that `event.context.player` always refers to the player who is considered attacking.

tvdstaaij:
Jadiel pointed out that the ruling on Isawa Kaede applies here:

>If Akodo Toturi or Doji Hotaru win the conflict on defense and trigger their ability, the attacking player is considered to be resolving the ring effect, so Isawa Kaede would not be immune to it.
>-- Developer ruling

So the current behavior is correct. Good news for all Lion players with a big brain: this allows chaining Toturi and Akodo Motivator on earth defense, making your opponent draw 1 and discard 1 to get rid of that one card you don't want them to have, and discarding 1 yourself to turn on your "less cards in hand than opponent" effects!

server/game/cards/10-TEL/ArdentOmoidasu.js
jeremylarner: Akodo Motivator is a good model to follow here, I think :)

PR #3552, opened 2019-10-10
TEL - Insolent Outcast

### LINKED ISSUE #3509: TEL - Insolent Outcast
image

tvdstaaij:
closes #3509


tvdstaaij:
If anyone knows I was wondering why the test works the way I ended up doing it (with Way and SBR), but doesn't work in these ways:
* By calling `.honor()` on the character directly instead of through an honor effect card
* By declaring a hand like `['way-of-the-crane', 'way-of-the-crane']` and using the card two times (the second honor effect seems not to happen)

PR #3585, opened 2019-10-16
TEL - Akodo Motivator

### LINKED ISSUE #3519: TEL - Akodo Motivator
image

tvdstaaij:
closes #3519

Due to some confusion on my part over a few rule and implementation details I wanted to cover a lot of angles in the tests, so they may be a bit long-winded. Well, I guess extensive tests never do much harm :)

I should note that the tests also pass without the check for isOpponentsEffect; I think this is because onCardsDiscardedFromHand doesn't fire on discard effects from friendly cards in the first place? I left this in for now, but if this is by design and the check is completely redundant give a shout and I'll take it out.


server/game/cards/10-TEL/AkodoMotivator.js
```diff
@@ -0,0 +1,27 @@
...
+const AbilityDsl = require('../../abilitydsl');
+
+class AkodoMotivator extends DrawCard {
+    setupCardAbilities() {
+        this.reaction({
+            title: 'Opponent discards an equal number of cards at random',
+            when: {
+                onCardsDiscardedFromHand: (event, context) => {
+                    const discardedFromOwnHand = (event.player === context.player);
+                    const discardedByOpponentsEffect = (event.player === event.context.player.opponent);
```
jeremylarner: I would prefer this to be `event.player.opponent === event.context.player`
`event.player` always exists but `event.context.player` doesn't always exist

server/game/cards/10-TEL/AkodoMotivator.js
```diff
@@ -0,0 +1,27 @@
...
+                    const discardedByRingEffect = (event.context.source.type === 'ring');
+                    const discardedByCardEffect = event.context.ability.isCardAbility();
+                    return discardedFromOwnHand && discardedByOpponentsEffect && (discardedByRingEffect || discardedByCardEffect);
+                }
+            },
+            gameAction: AbilityDsl.actions.discardAtRandom(context => ({
+                amount: _.isArray(context.event.cards) ? context.event.cards.length : context.event.amount
```
jeremylarner: This should be just `context.event.amount`, but I suspect there is a reason why it isn't.... I think we need to fix the reason though, not hack around it :)

server/game/cards/10-TEL/AkodoMotivator.js
tvdstaaij: Done.

server/game/cards/10-TEL/AkodoMotivator.js
tvdstaaij: Yeah the reason is that `amount` doesn't exist when the discard is chosen, and `cards` doesn't exist when the discard is random. I agree that it would be good to make it so that amount (and also cards? could be more complex depending on the current implementation) is always available. I could take a look at that, it doesn't sound like it would be a very difficult change for a newcomer.

jeremylarner: Yeah, amount should be on the event when the discard is chosen, that's the problem.  If you can find your way round game actions and add that, that would be great.

tvdstaaij: Fixed.

jeremylarner:
Same here - can you re-PR this with just the card and test?

jeremylarner:
Resubmitted as #3641

PR #3588, opened 2019-10-16
aCf - Unified company

### LINKED ISSUE #3581: aCF - Unified Company
image

Matthew-Griffith:
closes #3581

### DIFF
--- a/test/server/cards/09.5-aCF/UnifiedCompany.spec.js
+++ b/test/server/cards/09.5-aCF/UnifiedCompany.spec.js
@@ -1,0 +1,71 @@
+describe('Unified Company', function() {
+    integration(function() {
+        describe('Unified Company\'s ability', function() {
+            beforeEach(function() {
+                this.setupTest({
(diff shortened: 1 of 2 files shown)


test/server/cards/09.5-aCF/UnifiedCompany.spec.js
```diff
@@ -0,0 +1,71 @@
...
+            it('should put the character into play', function() {
+                expect(this.gunso.location).toBe('dynasty discard pile');
+                this.noMoreActions();
+                this.player1.clickCard(this.unifiedCompany);
+                this.player1.clickCard(this.gunso);
+                expect(this.gunso.location).toBe('play area');
+                expect(this.gunso.isParticipating()).toBe(false);
+            });
+
+            it('should not trigger if you have equal or less cards in hand', function() {
```
marcsoiferman-cm: I would split this test case into "should not trigger if you have equal cards in hand" and "should not trigger if you have more cards in hand".

PR #3596, opened 2019-10-18
Bugfix - Handle undefined skill stats 

jamesoddy:
Quick fix for handling undefined skill stats (as when stored on mongodb on live)


NoahTheDuke:
This is a bit hard to read. Might it be helpful to pull one or both levels out into separate `if` statements?

PR #3601, opened 2019-10-19
aCF - Withstand the Darkness (Do Not Merge)

marcsoiferman-cm:
This is a disaster, but the tests are useful


marcsoiferman-cm:
Needs #3622

PR #3604, opened 2019-10-20
aCF - Shukujo

### LINKED ISSUE #3578: aCF - Shukujo
image

marcsoiferman-cm:
Closes #3578

Chat logs might need a bit of work?  Only displays "[x] uses the gained ability from Shukujo", but doesn't say that it switched the conflict type.

### DIFF
--- a/server/game/cards/09.5-aCF/Shukujo.js
+++ b/server/game/cards/09.5-aCF/Shukujo.js
@@ -3,7 +3,7 @@ const AbilityDsl = require('../../abilitydsl');
 const { AbilityTypes } = require('../../Constants');
 
 class Shukujo extends DrawCard {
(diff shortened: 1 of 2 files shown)


server/game/cards/09.5-aCF/Shukujo.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    canAttach(card, context) {
+        if(card.controller !== context.player) {
+            return false;
+        }
+        return card.isUnique() && card.isFaction('crane') ? super.canAttach(card, context) : false;
+    }
+}
+
```
The-Outcast: You should probably use the new this.attachmentConditions.

Issue #3574, opened 2019-10-15
aCF - Palace Guard


Matthew-Griffith:
![image](https://user-images.githubusercontent.com/22338548/66855083-95228380-ef50-11e9-95d8-8f24139c62cd.png)

marcsoiferman-cm:
I'll do this one

PR #3606, opened 2019-10-20
aCF - Palace Guard

marcsoiferman-cm:
Implementation and tests for #3574


marcsoiferman-cm:
Closed - selected develop instead of aCF

Issue #3502, opened 2019-09-25
TEL - Kitsuki Kagi


The-Outcast:
![image](https://user-images.githubusercontent.com/50174301/65646400-46419800-dffb-11e9-946f-f442dc8d4aec.png)

marcsoiferman-cm:
Working on this

PR #3611, opened 2019-10-22
TEL - Kitsuki Kagi

marcsoiferman-cm:
Implementation and tests for #3502.


marcsoiferman-cm:
Had to update RemoveFromGameAction.ts to make this work.  Not sure if the update is okay.

jeremylarner:
Yeah, there are a couple of issues:
1) This seems a really awkward way to implement targeting.  Really we should have a selector which works for this, but we don't so I don't think this is a problem
2) The idea behind `location` in RemoveFromGame is so that the engine can track whether the card moves zone between the effect starting to resolve and finally resolving.  `location` should be where the card started.  Having written this, I wonder if we're really doing this right, and maybe `location` shouldn't be a property at all...

jeremylarner:
Now I look into it, I realise you shouldn't be using the removeFromGame action at all.  That's only for cards which are currently in play.  You should be using moveCard instead.

Issue #3573, opened 2019-10-15
aCF - Outflank


Matthew-Griffith:
![image](https://user-images.githubusercontent.com/22338548/66855064-8a67ee80-ef50-11e9-82c4-d1475e321583.png)

marcsoiferman-cm:
Going to try this one

PR #3613, opened 2019-10-22
aCF - Outflank

marcsoiferman-cm:
Testing and implementation for #3573.

Currently fails a test - allows you to select the same character twice with multiple copies of Outflank.  Looking for feedback on if this is acceptable or not.

### DIFF
--- a/test/server/cards/09.5-aCF/Outflank.spec.js
+++ b/test/server/cards/09.5-aCF/Outflank.spec.js
@@ -6,44 +6,57 @@ describe('Outflank', function() {
                     phase: 'conflict',
                     player1: {
                         inPlay: ['moto-youth'],
-                        hand: ['outflank', 'chasing-the-sun']
+                        hand: ['outflank', 'chasing-the-sun'],
+                        conflictDiscard: ['outflank']
                     },
                     player2: {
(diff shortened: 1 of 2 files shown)


marcsoiferman-cm:
Failing a test - you're able to select the same character twice with two copies of Outflank.  This probably isn't okay, but I don't know how to check if a lasting effect of `cardCannot('declareAsDefender')` is already on a character.

jeremylarner:
That's fine - you can Outflank the same character twice.

test/server/cards/09.5-aCF/Outflank.spec.js
```diff
@@ -0,0 +1,193 @@
...
+                this.noMoreActions();
+                this.initiateConflict({
+                    attackers: [this.motoYouth],
+                    province: this.toshiRanbo
+                });
+                expect(this.player1).not.toHavePrompt('Triggered Abilities');
+                expect(this.player2).toHavePrompt('Choose Defenders');
+            });
+
+            it('should not trigger when a province is revealed mid-conflict', function() {
```
jeremylarner: I think this should trigger mid-conflict also....

Issue #3570, opened 2019-10-15
aCF - Frontline Engineer


Matthew-Griffith:
![image](https://user-images.githubusercontent.com/22338548/66854948-58568c80-ef50-11e9-9b8a-a26bb2dc8ab2.png)

marcsoiferman-cm:
Implemented this one

PR #3627, opened 2019-10-30
aCF - Frontline Engineer

marcsoiferman-cm:
Implements #3570

### DIFF
--- a/server/game/cards/09.5-aCF/FrontlineEngineer.js
+++ b/server/game/cards/09.5-aCF/FrontlineEngineer.js
@@ -1,0 +1,39 @@
+const DrawCard = require('../../drawcard.js');
+const { Locations, Players, CardTypes } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl.js');
+
+class FrontlineEngineer extends DrawCard {
+    setupCardAbilities(ability) {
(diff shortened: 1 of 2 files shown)


server/game/cards/09.5-aCF/FrontlineEngineer.js
```diff
@@ -0,0 +1,44 @@
...
+            title: 'Place a holding from your deck faceup in the defending province',
+            condition: context => context.player.dynastyDeck.size() > 0 && context.player.isDefendingPlayer(),
+            effect: 'look at the top five cards of their dynasty deck',
+            handler: context => this.game.promptWithHandlerMenu(context.player, {
+                activePromptTitle: 'Choose a holding',
+                context: context,
+                cardCondition: card => card.getType() === CardTypes.Holding,
+                cards: context.player.dynastyDeck.first(5),
+                cardHandler: cardFromDeck => {
+                    let card = context.player.getDynastyCardInProvince(this.game.currentConflict.conflictProvince.location);
```
The-Outcast: Frontline Engineer mentions "discard each other card in the province" so this should probably be getDynasty**Cards**InProvince. As there are cards which allow for multiple cards to be present in a province, like Backalley Hideout.

Issue #3632, opened 2019-11-04
Emperor's Summons places the chosen card facedown


The-Outcast:

Syrail:
Fix merged. This can be closed.

PR #3644, opened 2019-11-09
Fix issue #3632 - Emperor's Summons Places the chosen card facedown

Syrail:
Fix for issue#3632, Test added to verify card is faceup.


The-Outcast:
closes #3632

Issue #3483, opened 2019-09-15
Sharpen the Mind discards attached character instead of prompting a card?


The-Outcast:
https://media.discordapp.net/attachments/570280058103529474/622932407292526612/SharpenTheMindglitch.png?width=926&height=509

Syrail:
This can be closed now as well.

PR #3646, opened 2019-11-10
Fix #3645: Sharpen the mind filters discards incorrectly

### LINKED ISSUE #3645: Sharpen the Mind improperly filters discard targets
Sharpen the Mind lets you choose cards from any area, including out of play areas, to discard for its cost. This should be restricted to player's hand only. (Root cause of #3483 )

Syrail:
Added test case to reproduce error behavior.
Updated Sharpen The Mind implementation to only permit choosing from hand.


Syrail:
Fixed typo in title - This PR addresses #3646 and  #3483

The-Outcast:
closes  #3646, #3483

Issue #3638, opened 2019-11-06
Overhear's moved and chat stated cards don't match


The-Outcast:
https://cdn.discordapp.com/attachments/570280058103529474/641588092977020928/unknown.png
![image](https://user-images.githubusercontent.com/50174301/68295070-d6aad800-0091-11ea-8ef2-759d7020c5a6.png)

Syrail:
Fix merged. This can be safely closed.

PR #3647, opened 2019-11-10
Fix for #3638 - Overhear logs wrong card

Syrail:
Resolves Issue #3638
Overhear now sequentially moves a random card to top of opponent's deck, then uses a LookAt action to reveal said card.


The-Outcast:
closes #3638

PR #3688, opened 2019-11-16
DoR - Northern Curtain Wall

### LINKED ISSUE #3668: DoR - Northern Curtain Wall

marcsoiferman-cm:
Closes #3668


marcsoiferman-cm:
Looked at this again, it won't work with set effects.

Misread how the other skill constructors work - part of getBaseModifiers grabs the printed value.

PR #3699, opened 2019-11-20
Disguised and Conditional Attachments

The-Outcast:
Add a test for disguised and attachments with attachmentconditions that are false.


gryffon:
Closing in preparation of migration. Can be re-opened against new repository after migration if necessary.

PR #3707, opened 2019-11-21
DoR - Strength of the Mountain

das3in:


server/game/cards/11-DoR/TheStrengthOfTheMountain.js
```diff
@@ -0,0 +1,27 @@
...
+            effect: 'prevent defending characters from bowing or returning home while defending',
+            gameAction: [
+                ability.actions.cardLastingEffect(context => ({
+                    target: context.player.cardsInPlay.filter(card => card.isDefending()),
+                    effect: ability.effects.doesNotBow()
+                })),
+                ability.actions.cardLastingEffect(context => ({
+                    target: context.player.cardsInPlay.filter(card => card.isDefending()),
+                    effect: ability.effects.cardCannot({
+                        cannot: 'returnHome'
```
marcsoiferman-cm: I think this should be 'sendHome' not 'returnHome'.

You're also missing the qualifier `restricts: 'opponentsCardEffects'` - card as written should allow you to use your own cards to move defenders home.

server/game/cards/11-DoR/TheStrengthOfTheMountain.js
```diff
@@ -0,0 +1,27 @@
+const DrawCard = require('../../drawcard.js');
+
+class TheStrengthOfTheMountain extends DrawCard {
+    setupCardAbilities(ability) {
```
marcsoiferman-cm: From my understanding the `ability` parameter has been deprecated.  Include AbilityDsl (`const AbilityDsl = require('../../abilitydsl');`) and use `AbilityDsl.` instead of `ability`.

server/game/cards/11-DoR/TheStrengthOfTheMountain.js
```diff
@@ -0,0 +1,27 @@
...
+                    })
+                }))
+            ]
+        });
+    }
+}
+
+TheStrengthOfTheMountain.id = 'the-strength-of-the-mountain';
+
+module.exports = TheStrengthOfTheMountain;
```
marcsoiferman-cm: You're also missing the "each defending character cannot be bowed" bit (you've only done "each defending character cannot be ... moved home by opponents' card effects" and "each defending character .... does not bow as the result of conflict resolution.")

The-Outcast:
Closed due to inactivity and has been picked up in #3743

PR #3720, opened 2019-11-24
DoR - Kaiu Shihobu (using RFG - Don't merge)

### LINKED ISSUE #3661: DoR - Kaiu Shihobu

marcsoiferman-cm:
Closes #3661

Uses Removed from Game, and the card selector menu is a bit big.  But works decently well.  Holdings removed from game are revealed (dumped in the chat log), but are facedown in the pile for the opponent.

Going to work on adding support for cards under other cards, using the following as reference:
UI - https://github.com/throneteki/throneteki-client/pull/23/files
Backend - https://github.com/throneteki/throneteki/pull/2078/files
(For Reference: https://github.com/throneteki/throneteki/blob/master/server/game/cards/11.2-TMoW/Meereen.js)


marcsoiferman-cm:
Redundant with 3768

PR #3774, opened 2019-12-07
SD - Inferno Guard Invoker

### LINKED ISSUE #3758: SD - Inferno Guard Invoker
image

The-Outcast:
closes #3758

### DIFF
--- a/server/game/cards/09.6-SD/InfernoGuardInvoker.js
+++ b/server/game/cards/09.6-SD/InfernoGuardInvoker.js
@@ -1,9 +1,15 @@
+const { EventNames }  = require('../../Constants.js');
+const EventRegistrar = require('../../eventregistrar.js');
 const DrawCard = require('../../drawcard.js');
-const { CardTypes, Durations, Players, Locations } = require('../../Constants');
+const { CardTypes, Durations, Players } = require('../../Constants');
 const AbilityDsl = require('../../abilitydsl');
 
 class InfernoGuardInvoker extends DrawCard {
     setupCardAbilities() {
+        this.provinceBroken = false;
+        this.eventRegistrar = new EventRegistrar(this.game, this);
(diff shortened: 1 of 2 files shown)


test/server/cards/09.6-SD/InfernoGuardInvoker.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+                expect(this.getChatLogs(5)).not.toContain('Serene Warrior is discarded, burned to a pile of ash due to the delayed effect of Inferno Guard Invoker');
+                this.player1.clickPrompt('Don\'t Resolve');
+
+                expect(this.sereneWarrior.location).toBe('play area');
+                expect(this.player1).toHavePrompt('Action Window');
+            });
+        });
+    });
+});
```
marcsoiferman-cm: Can you add a test for this ruling?

If a second delayed effect occurs at the end of the conflict (eg. Forebearer's Echoes), and it also causes a character to leave play, the first player will decide which resolves first. The second will not resolve, as the card has already left play.

server/game/cards/09.6-SD/InfernoGuardInvoker.js
```diff
@@ -0,0 +1,37 @@
...
+            },
+            gameAction: AbilityDsl.actions.multiple([
+                AbilityDsl.actions.honor(context => ({ target: context.target})),
+                AbilityDsl.actions.cardLastingEffect(context => ({
+                    target: context.target,
+                    duration: Durations.UntilEndOfConflict,
+                    location: [Locations.DynastyDiscardPile, Locations.PlayArea],
+                    effect: AbilityDsl.effects.delayedEffect({
+                        when: {
+                            onBreakProvince: () => true
```
marcsoiferman-cm: Based on the card text I think should be onConflictFinished()?

PR #3777, opened 2019-12-07
Develop - Know the Terrain

### LINKED ISSUE #3760: SD - Know the Terrain
image

marcsoiferman-cm:
Closes #3760

Needs a ruling about when exactly in the conflict declaration the interrupt window for this opens.


marcsoiferman-cm:
Needs testing to see what happens if all attackers becomes ineligible after this event is played but before the conflict officially starts

server/game/cards/09.6-SD/KnowTheTerrain.js
```diff
@@ -0,0 +1,37 @@
...
+class KnowTheTerrain extends DrawCard {
+    setupCardAbilities() {
+        this.interrupt({
+            title: 'Switch the attacked province with a facedown province',
+            when: {
+                onConflictDeclaredBeforeReveal: (event, context) => event.conflict.conflictProvince.facedown && 
+                    context.player.isDefendingPlayer() &&
+                    event.conflict.conflictProvince.location !== Locations.StrongholdProvince
+            },
+            target: {
```
jeremylarner: if the card doesn't target, you need to use a selectCard action (or promptForSelect if you're using a handler)

server/game/player.js
```diff
@@ -1177,6 +1177,45 @@ class Player extends GameObject {
         return this.anyEffect(EffectNames.EventsCannotBeCancelled);
     }
 
+    switchProvinces(province1, province2) {
```
jeremylarner: This should be dealt with by `moveCard`

server/game/player.js
```diff
@@ -1177,6 +1177,45 @@ class Player extends GameObject {
...
+
+        let index1 = cardsInLocation1.indexOf(province1);
+        let index2 = cardsInLocation2.indexOf(province2);
+        cardsInLocation1.splice(index1, 1);
+        cardsInLocation2.splice(index2, 1);
+        cardsInLocation1.push(province2);
+        cardsInLocation2.push(province1);
+    }
+
+    getProvinceArray(location) {
```
jeremylarner: This should be dealt with by `getSourceList`

server/game/gamesteps/conflict/conflictflow.js
```diff
@@ -171,6 +172,23 @@ class ConflictFlow extends BaseStepWithPipeline {
...
+        let ring = this.conflict.ring;
+        let events = [this.game.getEvent(EventNames.OnConflictDeclaredBeforeReveal, {
+            conflict: this.conflict,
+            type: this.conflict.conflictType,
+            ring: ring,
+            attackers: this.conflict.attackers.slice(),
+            ringFate: ring.fate
+        })];
+
+        this.game.openEventWindow(events);
```
jeremylarner: This isn't really the right way to do this - we need to make conflict declaration a proper event which can be interrupted and reacted to.  We need to know where covert fits in to do it correctly though

marcsoiferman-cm:
Not sure how important this ruling actually is.  The event works as expected now, based on the RAW.  Updates for this required splitting taking ring fate and revealing the province, so the chat messages are re-arranged, I think.

gryffon:
Closing in preparation of migration. Can be re-opened against new repository after migration if necessary.

PR #3847, opened 2019-12-21
SoW - Iron Foundations Stance

### LINKED ISSUE #3813: SoW - Iron Foundations Stance
image

marcsoiferman-cm:
Closes #3813


server/game/cards/12-SoW/IronFoundationsStance.js
```diff
@@ -0,0 +1,45 @@
...
+                return {
+                    gameAction: AbilityDsl.actions.draw(),
+                    message: '{0} draws a card'
+                };
+            },
+            effect: 'prevent opponents\' actions from bowing or moving home {0}'
+        });
+    }
+}
+
```
The-Outcast: Doesn't the then make it dependent on the previous actions? Both are lasting effects so they probably always happen, but shouldn't this be a AbilityDsl.actions.conditional action grouped together with the other 2?

server/game/cards/12-SoW/IronFoundationsStance.js
```diff
@@ -0,0 +1,45 @@
+import { CardTypes, Players } from '../../Constants.js';
+const DrawCard = require('../../drawcard.js');
+const AbilityDsl = require('../../abilitydsl.js');
+
+class IronFoundationsStance extends DrawCard {
+    setupCardAbilities() {
+        this.action({
+            title: 'Lower military skill',
```
The-Outcast: Watch out when copying ;)

jeremylarner:
This should really use `this.game.currentConflict.getNumberOfCardsPlayed()` - if there's an issue with the way that function is counting played cards, we should fix that, not hack the implementation...

PR #3849, opened 2019-12-22
SoW - Iuchi Rimei

### LINKED ISSUE #3814: SoW - Iuchi Rimei
image

marcsoiferman-cm:
Closes #3814

image

### DIFF
--- a/server/game/basecard.ts
+++ b/server/game/basecard.ts
@@ -609,7 +609,7 @@ class BaseCard extends EffectSource {
         if(!parent || parent.getType() !== CardTypes.Character || !ignoreType && this.getType() !== CardTypes.Attachment) {
             return false;
         }
-        if(this.anyEffect(EffectNames.AttachmentMyControlOnly) && context.player !== parent.controller) {
+        if(this.anyEffect(EffectNames.AttachmentMyControlOnly) && context.player !== parent.controller && this.controller !== parent.controller) {
             return false;
         } else if(this.anyEffect(EffectNames.AttachmentUniqueRestriction) && !parent.isUnique()) {
             return false;
--- a/server/game/cards/12-SoW/IuchiRimei.js
+++ b/server/game/cards/12-SoW/IuchiRimei.js
@@ -1,0 +1,32 @@
+const DrawCard = require('../../drawcard.js');
+const { Players, CardTypes } = require('../../Constants');
+const AbilityDsl = require('../../abilitydsl.js');
+
+class IuchiRimei extends DrawCard {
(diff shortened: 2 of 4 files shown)


jeremylarner:
I don't think this needs a ruling - it cannot target attachments not attached to characters, as there's no potential to change the game state.  There's no reason whatsoever to think it would ever discard an attachment

server/game/basecard.ts
```diff
@@ -609,7 +609,7 @@ class BaseCard extends EffectSource {
         if(!parent || parent.getType() !== CardTypes.Character || !ignoreType && this.getType() !== CardTypes.Attachment) {
             return false;
         }
-        if(this.anyEffect(EffectNames.AttachmentMyControlOnly) && context.player !== parent.controller) {
+        if(this.anyEffect(EffectNames.AttachmentMyControlOnly) && context.player !== parent.controller && this.controller !== parent.controller) {
```
jeremylarner: I don't like that we compare `context.player` in one test here and `this.controller` in the other.  They should both be the same...

server/game/cards/12-SoW/IuchiRimei.js
```diff
@@ -0,0 +1,32 @@
...
+            title: 'Move an attachment',
+            target: {
+                cardType: CardTypes.Attachment,
+                controller: Players.Opponent,
+                cardCondition: card => card.costLessThan(2),
+                gameAction: AbilityDsl.actions.selectCard(context => ({
+                    cardCondition: card => card !== context.target.parent && card.controller === context.target.parent.controller && card.type === CardTypes.Character,
+                    message: '{0} moves {1} to {2}',
+                    messageArgs: card => [context.player, context.target, card],
+                    gameAction: AbilityDsl.actions.ifAble(context => ({
```
jeremylarner: this should just be an `attach` action

server/game/basecard.ts
marcsoiferman-cm: Any suggestions on how to use the same one for both?  Using Rimei, `context.player` is player1, `parent.controller` is player2, and `this.controller` is player2.

It looks like steal effects (ex: Karada) are relying on the `context.player !== parent.controller` bit.  Not sure why.

server/game/cards/12-SoW/IuchiRimei.js
marcsoiferman-cm: Removing the `ifAble` bit causes you to not be able to move attachments that would end up being illegal.  If I make this just `attach`, then the Force of the River test fails.

PR #3867, opened 2019-12-31
Apply new RRG changes around discard pile order

The-Outcast:
Has functional impact on plagerist

### DIFF
--- a/server/game/cards/02.6-MotE/IllustriousPlagiarist.js
+++ b/server/game/cards/02.6-MotE/IllustriousPlagiarist.js
@@ -5,7 +5,7 @@ class IllustriousPlagiarist extends DrawCard {
     setupCardAbilities(ability) {
         this.action({
             title: 'Copy action abilty of opponent\'s top event',
-            condition: context => context.player.opponent,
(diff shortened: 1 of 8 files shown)


server/game/cards/02.6-MotE/IllustriousPlagiarist.js
```diff
@@ -1,16 +1,19 @@
...
             target: {
+                player: Players.Opponent, // As per December 2019 RRG Conflict Discard Pile order is determined by the controller of the pile
                 location: Locations.ConflictDiscardPile,
                 controller: Players.Opponent,
-                cardCondition: (card, context) => card === context.player.opponent.conflictDiscardPile.find(card => card.type === CardTypes.Event) &&
-                                                  card.abilities.actions.length > 0,
+                cardCondition: (card, context) => card.location === Locations.ConflictDiscardPile &&
+                    card.type === CardTypes.Event &&
+                    card.controller === context.player.opponent &&
+                    card.abilities.actions.length > 0,
```
marcsoiferman-cm: In theory this should let the opponent choose an event without an action.

"Choose the topmost event in an opponent's conflict discard pile – this character gains each of that event's action abilities until the end of the phase.".  If the topmost event has no actions, you should still be able to choose it and have the effect fizzle (due to failure to resolve).

The other option would be to change the condition to "opponent's discard pile contains no events without action abilities" so that you just can't trigger him if your opponent could render him blank.

PR #3869, opened 2019-12-31
Daidoji Netsu should stop cards being returned to deck/hand

jeremylarner:


marcsoiferman-cm:
Replaced with #3890 (same branch, but with added tests)

PR #3878, opened 2020-01-07
Updated the FoF effect and chat messages

marcsoiferman-cm:
Messages weren't updated with the errata


marcsoiferman-cm:
Duplicated by #3879

PR #3881, opened 2020-01-10
Bugfixes for Chukan Nobue, Tadaka & Investigator

### LINKED ISSUE #3684: Kitsuki Investigator menu inescapable with Chukan Nobue in play
Chukan Nobue's text that prevents opponents' card abilities from discarding cards causes Kitsuki Investigator to open an inescapable menu when his action is used.

marcsoiferman-cm:
Fixes #3684

### DIFF
--- a/server/game/cards/09.2-BoB/IsawaTadaka2.js
+++ b/server/game/cards/09.2-BoB/IsawaTadaka2.js
@@ -12,13 +12,18 @@ class IsawaTadaka2 extends DrawCard {
                 location: Locations.DynastyDiscardPile,
                 mode: TargetModes.Unlimited
             }),
-            gameAction: AbilityDsl.actions.cardMenu(context => ({
-                cards: context.player.opponent && context.costs.removeFromGame ?
-                    context.player.opponent.hand.shuffle().slice(0, context.costs.removeFromGame.length) :
-                    [context.source],
-                gameAction: AbilityDsl.actions.discardCard(),
-                message: '{0} chooses {1} to be discarded from {2}',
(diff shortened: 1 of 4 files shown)


server/game/cards/09.2-BoB/IsawaTadaka2.js
```diff
@@ -12,13 +12,18 @@ class IsawaTadaka2 extends DrawCard {
...
                 mode: TargetModes.Unlimited
             }),
-            gameAction: AbilityDsl.actions.cardMenu(context => ({
-                cards: context.player.opponent && context.costs.removeFromGame ?
-                    context.player.opponent.hand.shuffle().slice(0, context.costs.removeFromGame.length) :
-                    [context.source],
-                gameAction: AbilityDsl.actions.discardCard(),
-                message: '{0} chooses {1} to be discarded from {2}',
-                messageArgs: (card, player, cards) => [player, card, cards]
+            gameAction: AbilityDsl.actions.ifAble(() => ({
```
jeremylarner: I'm not sure card implementations should be expected to be able to deal with this explicitly.  I think the game engine should handle this case better

marcsoiferman-cm: so the issue with Tadaka is that we do the "look at" and "discard" in the same step; if the discard is illegal, the whole step is skipped and you don't get to look at any cards.  Is there a way to do this like Investigator where there's a "reveal" action first, which randomly selects the cards, and then does the discard action using the same cards as in the reveal action?  I split it up because I didn't know if you could pass the subarray to the next action.

